PR #8, opened 2014-11-05
KAA-129: Change to default schemas namespace and fix to issues with empt...

ashvayka:
...y default profile


server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/model/mongo/EndpointProfile.java
```diff
@@ -34,6 +34,7 @@
 import org.springframework.data.mongodb.core.mapping.Document;
 import org.springframework.data.mongodb.core.mapping.Field;
 
+import com.mongodb.BasicDBObject;
```
ikulikov: Unused import.

PR #10, opened 2014-11-06
KAA-164: Fix source templates used for C++ SDK generation.

ghost:


server/control/src/main/resources/sdk/cpp/event/EventFamily.hpp.template
```diff
@@ -17,13 +17,15 @@
...
+#ifdef KAA_USE_EVENTS
+
 #include <set>
 #include <list>
 #include <string>
 #include <vector>
 #include <sstream>
 
-#include <boost/cstdint.hpp>
+#include <cstdint.hpp>
```
izeigerman: Use "<cstdint>" instead

ghost:
LGTM

PR #25, opened 2014-11-24
KAA-212: Fix resetT enum adding in schema generation algorithm

izeigerman:


common/core/src/main/java/org/kaaproject/kaa/server/common/core/algorithms/schema/SchemaCreatorImpl.java
```diff
@@ -159,7 +159,7 @@ private Boolean isAddressableValue(Object value) {
 
     private void checkIfArray(Map<String, Object> fieldType, List<Object> union) {
```
ashvayka: Name of this method seems inconsistent. If method is named "checkIfArray", one assume that it returns boolean with result of a check. This method is changing state of an input parameter, so it definitely should be renamed.

PR #36, opened 2014-12-03
KAA-240: Invalidate sandbox and admin SDK caches on change services host/ip operation

ikulikov:


server/admin/src/main/java/org/kaaproject/kaa/server/admin/controller/KaaAdminController.java
```diff
@@ -372,18 +372,33 @@ public void getSdk(@RequestBody SdkKey key,
...
+     * Flushes all cached Sdks within tenant.
+     *
+     */
+    @RequestMapping(value="flushSdkCache", method=RequestMethod.POST)
+    @ResponseStatus(value = HttpStatus.OK)
+    public void flushSdkCache() throws KaaAdminServiceException {
+        try {
+            kaaAdminService.flushSdkCache();
+        }
+        catch (Exception e) {
```
ghost: According to the code style, the enclosing brace of the 'try' block should be on one line with the 'catch' statement.

server/admin/src/main/java/org/kaaproject/kaa/server/admin/shared/file/FileData.java
```diff
@@ -0,0 +1,56 @@
+package org.kaaproject.kaa.server.admin.shared.file;
```
ghost: The license header should be put into the begin of the source file.

shpigor:
Reviewed

PR #42, opened 2014-12-10
KAA-254: Add Sandbox Web REST API to manage build of demo projects. Add automatic build of demo projects via Sandbox REST API to sandbox builder.

ikulikov:


shpigor:
Reviewed

PR #43, opened 2014-12-11
KAA-267: Add support of new 'Generic' flume event format to Flume appender.

ikulikov:


ashvayka:
LGTM

server/appenders/flume/appender/src/main/java/org/kaaproject/kaa/server/appenders/flume/appender/FlumeAvroEventBuilder.java
```diff
@@ -71,5 +100,14 @@ public Event generateEvent(String appToken, int schemaVersion, List<LogEvent> lo
         LOG.trace("Build flume event with array body [{}]", baos);
         return event;
     }
-
+    
+    private List<Event> generateGenericEvent(LogSchema schema, List<LogEvent> logEvents) {
```
shpigor, line 104: Is it possible situation where logEvents equals null? If yes, need to check it for null.

shpigor:
Reviwed

PR #64, opened 2015-01-09
KAA-299: Implement an alternative option that would allow the user to populate the notification message data by filling in a form generated by avro-ui.

ikulikov:


server/common/server-shared/src/main/java/org/kaaproject/kaa/server/common/Version.java
```diff
@@ -27,10 +27,10 @@
 public final class Version {
 
     /** The Constant PROJECT_VERSION. */
-    public static final String PROJECT_VERSION = "N/A";
+    public static final String PROJECT_VERSION = "0.7.0-SNAPSHOT";
```
ashvayka: No need to change this. Please revert this changes or add to .gitignore

ashvayka:
reviewed

server/admin/src/main/java/org/kaaproject/kaa/server/admin/client/mvp/place/SendNotificationPlace.java
```diff
@@ -0,0 +1,110 @@
...
+        return true;
+    }
+
+    @Override
+    public boolean isLeaf() {
+        return true;
+    }
+
+    @Override
+    public TreePlaceDataProvider getDataProvider(EventBus eventBus) {
```
shpigor: I think it's better to move this implementation to abstract class TreePlace, and override it when needed.

shpigor:
Reviewed

PR #70, opened 2015-01-19
KAA-284: Add the Bootstrap extension to the binary platform layer protocol.

ghost:


client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,283 @@
...
+    kaa_transport_channel_interface_t *channel = (kaa_transport_channel_interface_t *)data;
+    kaa_transport_channel_interface_t *matcher = (kaa_transport_channel_interface_t *)context;
+
+    error_code = channel->get_protocol_info(channel->context, &info1);
+    KAA_RETURN_IF_ERR(error_code);
+    error_code = matcher->get_protocol_info(matcher->context, &info2);
+    KAA_RETURN_IF_ERR(error_code);
+
+    return (0 == memcmp(&info1, &info2, sizeof(kaa_transport_protocol_info_t)));
+}
```
izeigerman, line 91: I think matching by protocol info is not a good idea. What if I add two KaaTCP channels (one channel for connection with bootstrap server and the other one for connection with operation server) at the same time?

client/client-multi/client-c/src/platform/ext_transport_channel.h
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SRC_PLATFORM_EXT_TRANSPORT_CHANNEL_H_
+#define SRC_PLATFORM_EXT_TRANSPORT_CHANNEL_H_
```
izeigerman: The guard macro should include only file name not the full path. And the file's description is missing here.

client/client-multi/client-c/src/platform/ext_transport_channel.h
```diff
@@ -0,0 +1,92 @@
...
+
+#include "kaa_common.h"
+
+# ifdef __cplusplus
+extern "C" {
+# endif
```
izeigerman: Why do we need spaces here?

client/client-multi/client-c/src/platform/ext_transport_channel.h
```diff
@@ -0,0 +1,92 @@
...
+    void                              *context;
+    kaa_get_protocol_info_fn          get_protocol_info;
+    kaa_get_supported_services_fn     get_supported_services;
+    kaa_sync_handler_fn               sync_handler;
+    kaa_release_channel_context_fn    release_context; /**< May be NULL */
+} kaa_transport_channel_interface_t;
+
+# ifdef __cplusplus
+}      /* extern "C" */
+# endif
```
izeigerman: Same problem with spaces here.

client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,283 @@
...
+    while (channel) {
+        error_code = channel->get_supported_services(channel->context
+                                                   , &services
+                                                   , &service_count);
+        if (error_code || !services || !service_count) {
+            KAA_LOG_WARN(self->logger, error_code, "Failed to retrieve list of supported services", error_code);
+            continue;
+        }
+
+        for (; service_count--;) {
```
izeigerman: Why don't you use a "while" loop here?

client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,283 @@
...
     }
 }
 
-kaa_error_t kaa_channel_manager_add_sync_handler(kaa_channel_manager_t *self
-        , kaa_sync_handler_fn handler, const kaa_service_t *supported_services, size_t services_count)
+kaa_error_t kaa_channel_manager_add_transport_channel(kaa_channel_manager_t *self
+                                                    , kaa_transport_channel_interface_t *channel)
 {
-    KAA_RETURN_IF_NIL4(self, handler, supported_services, services_count, KAA_ERR_BADPARAM);
+    KAA_RETURN_IF_NIL2(self, channel, KAA_ERR_BADPARAM);
```
yzeygerman: I believe we should return some channel's ID back to the user. This ID could be calculated using some hash function of callback pointers + context pointer. Also the same ID could be used by the user to remove his channel from the manager. And after all we will use this ID in "find_channel" function.

So the final structure that we are going to store in the list should look like following:
typedef struct {
    uint32_t channel_id;
    kaa_transport_channel_interface_t channel;
} kaa_transport_channel_instance_t;

client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,283 @@
...
-
-    while (details) {
-        if (details->sync_fn == handler) {
-            if (!kaa_list_remove_at(&self->sync_handlers, handlers, destroy_sync_details))
-                return KAA_ERR_BAD_STATE;
-            return KAA_ERR_NONE;
-        }
-        handlers = kaa_list_next(handlers);
-        details = kaa_list_get_data(handlers);
+    KAA_RETURN_IF_NIL2(self, channel, KAA_ERR_BADPARAM);
```
yzeygerman: I guess such "remove" API wouldn't work in practice. User has got only two ways to use it:
1) Store his "kaa_transport_channel_interface_t" instance in the heap or static memory.
2) User should create and fill "kaa_transport_channel_interface_t" when he's adding a channel and then create and fill the same instance when he's trying to remove it.

Both ways are not very usable IMHO.

client/client-multi/client-c/src/kaa_common.h
```diff
@@ -66,26 +66,35 @@ typedef enum {
     KAA_SERVICE_LOGGING = 4,
 } kaa_service_t;
 
+/**
+ * @brief Unique identifier used to represent client transport channel implementations.
+ */
+typedef struct {
+    uint32_t id;
+    uint16_t version;
+} kaa_transport_protocol_info_t;
```
yzeygerman: Is this structure represents the protocol ID or the unique channel ID? If it is an unique ID of the channel (not a transport protocol like KaaTCP) then my comments above are not fully correct. And if so then you should rename it to smth like "kaa_channel_info_t", because current name doesn't match to description. 
NOTE: Kaa SDK can hold several channel implementations with the same transport protocols.

client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,283 @@
...
     if (self) {
-        kaa_list_destroy(self->sync_handlers, destroy_sync_details);
+        kaa_list_destroy(self->transport_channels, destroy_channel);
         KAA_FREE(self);
     }
 }
 
-kaa_error_t kaa_channel_manager_add_sync_handler(kaa_channel_manager_t *self
-        , kaa_sync_handler_fn handler, const kaa_service_t *supported_services, size_t services_count)
+kaa_error_t kaa_channel_manager_add_transport_channel(kaa_channel_manager_t *self
```
izeigerman: I believe we should return some kind of channel's ID back to the user. This ID could be calculated using some hash function of callback pointers + context pointer. Also the same ID could be used by the user to remove his channel from the manager. And after all we will use this ID in "find_channel" function.

So the final structure that we are going to store in the list should look like following:
typedef struct {
    uint32_t channel_id;
    kaa_transport_channel_interface_t channel;
} kaa_transport_channel_instance_t;

client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,283 @@
...
+    self->sync_info.is_up_to_date = false;
+
+    KAA_LOG_INFO(self->logger, KAA_ERR_NONE, "New transport channel (id=0x%X, version=%u) added"
+                                                                        , protocol_info.id, protocol_info.version);
+
     return KAA_ERR_NONE;
 }
 
-kaa_error_t kaa_channel_manager_remove_sync_handler(kaa_channel_manager_t *self, kaa_sync_handler_fn handler)
+kaa_error_t kaa_channel_manager_remove_transport_channel(kaa_channel_manager_t *self
```
izeigerman: I guess such "remove" API wouldn't work in practice. User has got only two ways to use it:
1) Store his "kaa_transport_channel_interface_t" instance in the heap or static memory.
2) User should create and fill "kaa_transport_channel_interface_t" on stack when he's adding a channel and then create and fill the same instance when he's trying to remove it.

Both ways are not very usable IMHO.

client/client-multi/client-c/src/kaa_common.h
izeigerman: Does this structure represent the protocol ID or the unique channel ID? If it is an unique ID of the channel (not a transport protocol like KaaTCP) then my comments above are not fully correct. And if so then you should rename it to smth like "kaa_channel_info_t", because current name doesn't match to description. And please add a description for "id" and "version" fields.
NOTE: Kaa SDK is able to hold several channel implementations with the same transport protocol.

client/client-multi/client-c/src/platform/ext_transport_channel.h
ghost: Oops, autogenerated code.

client/client-multi/client-c/src/platform/ext_transport_channel.h
ghost: Copy-paste

client/client-multi/client-c/src/platform/ext_transport_channel.h
ghost: Copy-paste again

client/client-multi/client-c/src/kaa_channel_manager.c
ghost: No comments :) I will replace it with 'while'

client/client-multi/client-c/src/kaa_common.h
ghost: I agree, both documentation and struct name don't match to each other. This structure should represent unique identifier for a transport protocol.

client/client-multi/client-c/src/kaa_channel_manager.c
ghost, line 91: You are right, there is need to use some unique channel identifier. Thanks

client/client-multi/client-c/src/kaa_bootstrap_manager.h
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * @file kaa_bootstrap.h
+ * @brief Management of the Operations servers connection parameters.
+ *
+ * Manages connection parameters to Operations servers that are received from Bootstrap servers.
+ */
+
+#ifndef KAA_BOOTSTRAP_H_
+#define KAA_BOOTSTRAP_H_
```
izeigerman: Guard macro doesn't match to a file name.

client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,286 @@
...
 
-typedef struct kaa_sync_details {
-    kaa_sync_handler_fn sync_fn;
-    kaa_service_t      *supported_services;
-    size_t              supported_services_size;
-} kaa_sync_details;
+typedef struct {
+    uint32_t                             channel_id;
+    kaa_transport_channel_interface_t    channel;
+} kaa_transport_channel_wrapper_t;;
```
izeigerman: Extra semicolon.

client/client-multi/client-c/src/kaa_channel_manager.c
```diff
@@ -14,135 +14,286 @@
...
+
+kaa_error_t kaa_transport_channel_id_calculate(kaa_transport_channel_interface_t *channel
+                                             , uint32_t *channel_id)
+{
+    KAA_RETURN_IF_NIL2(channel, channel_id, KAA_ERR_BADPARAM);
+
+    const uint32_t prime = 31;
+
+    *channel_id = 1;
+    *channel_id = prime * *channel_id + (ptrdiff_t)channel->context;
```
izeigerman: Maybe it would be better to wrap *channel_id into brackets, like (*channel_id)

client/client-multi/client-c/src/collections/kaa_list.c
```diff
@@ -147,6 +147,30 @@ kaa_list_t *kaa_list_remove_at(kaa_list_t **head, kaa_list_t *position, dealloca
     return NULL;
 }
 
+kaa_error_t kaa_list_remove_first(kaa_list_t **head, match_predicate pred, void *context, deallocate_list_data deallocator)
+{
+    KAA_RETURN_IF_NIL2(head, pred, KAA_ERR_BADPARAM);
+
+    kaa_list_t *item_to_delete;
```
izeigerman: Please create variable only in that place where you're actually need it.

client/client-multi/client-c/src/kaa_bootstrap_manager.c
```diff
@@ -0,0 +1,313 @@
...
+    return access_point;
+}
+
+static kaa_error_t add_access_point(kaa_bootstrap_manager_t *self
+                                  , kaa_transport_protocol_id_t *protocol_info
+                                  , kaa_access_point_t *access_point)
+{
+    KAA_RETURN_IF_NIL2(protocol_info, access_point, KAA_ERR_BADPARAM);
+
+    kaa_channel_access_points_t *channel_access_points;
```
izeigerman: I guess there is no need to create "channel_access_points" right here.

andreypanasenko:
LGTM

izeigerman:
LGTM

PR #97, opened 2015-02-17
KAA-341: Server should reply with error code to invalid client messages

batytskyy:


server/transports/tcp/transport/src/test/java/org/kaaproject/kaa/server/transports/tcp/transport/TcpHandlerTest.java
```diff
@@ -112,9 +112,11 @@ public void testKaaSyncWithoutSession() throws Exception {
         UUID uuid = UUID.randomUUID();
         MessageHandler akkaService = Mockito.mock(MessageHandler.class);
         AbstractKaaTcpCommandProcessor msg = Mockito.mock(AbstractKaaTcpCommandProcessor.class);
-        Mockito.when(msg.getRequest()).thenReturn(new SyncRequest());
+        Mockito.when(msg.getRequest()).thenReturn(new KaaSync());
         TcpHandler handler = new TcpHandler(uuid, akkaService);
-        handler.channelRead0(null, msg);
+        ChannelHandlerContext context = buildDummyCtxMock();
+        handler.channelRead0(context, msg);
+        Mockito.verify(context).writeAndFlush(Mockito.any(ConnAck.class));
```
ashvayka: please add check that this is ConnAck with error code. Otherwise this may hide issues.

ashvayka:
Minor comment to the tests, otherwise LGTM

PR #99, opened 2015-02-17
KAA-336: fix bootstrap's server outdated removal

batytskyy:


server/common/zk/src/main/java/org/kaaproject/kaa/server/common/zk/WorkerNodeTracker.java
```diff
@@ -216,8 +223,16 @@ protected void endpointUpdated(ChildData data) {
...
+        String endpointAdress = constructEndpointAdress(nodeInfo);
+        Long removeTime = nodeInfo.getTimeStarted();
+        Long updateTime = operationNodesStartTimes.get(endpointAdress);
+        if (updateTime == null || removeTime >= updateTime) {
+            operationNodesStartTimes.remove(endpointAdress);
+            for (OperationsNodeListener listener : endpointListeners) {
+                listener.onNodeRemoved(nodeInfo);
+            }
+        } else {
+            LOG.debug("Ignoring [{}] endpoint removal, as it was before add/update: ", endpointAdress);
```
ashvayka: no need for  ":" in the end of the message

ashvayka:
Same logic should be implemented for bootstrap node list

PR #113, opened 2015-02-23
Kaa-387: Client receives notification from optional topic after successful unsubscription

ashvayka:


server/operations/src/main/java/org/kaaproject/kaa/server/operations/service/akka/actors/core/TopicActor.java
```diff
@@ -152,6 +155,16 @@ private void processEndpointRegistration(TopicRegistrationRequestMessage message
             context().watch(endpointActor);
         }
     }
+    
+
+    private void processEndpointDeregistration(TopicUnsubscriptionMessage message) {
+        String endpointKey = message.getOriginator().path().name();
+        if(endpointSessions.remove(endpointKey) != null){
```
shpigor: Code format

PR #116, opened 2015-02-25
KAA-371: Add support of floating-point numbers to the Avro C generator.

ghost:


client/client-multi/client-c/src/gen/kaa_logging_gen.c
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include <stddef.h>
```
andreypanasenko, line 16: I'm not sure, needs to check Econais compatibility

ghost, line 16: This source file is auto-generated, so if the deleted "include" is needed due the compatibility reason, the right place for it is a source template.

izeigerman:
LGTM

PR #121, opened 2015-02-27
KAA-397: Update the C++ SDK to support user verifiers.

ghost:


client/client-multi/client-cpp/impl/event/registration/EndpointRegistrationManager.cpp
```diff
@@ -284,25 +284,43 @@ void EndpointRegistrationManager::attachUser(const std::string& userExternalId
...
+    request->userExternalId = userExternalId;
+    request->userVerifierId = userVerifierToken;
+
+    if (listener != nullptr) {
+        attachStatusListener_ = listener;
+    }
+
+    userAttachRequest_.reset(request);
+
+    if (userTransport_ != nullptr) {
```
izeigerman: if (userTransport_)  would be enough.

client/client-multi/client-cpp/impl/event/registration/EndpointRegistrationManager.cpp
```diff
@@ -284,25 +284,43 @@ void EndpointRegistrationManager::attachUser(const std::string& userExternalId
...
-            userTransport_->sync();
-        } else {
-            KAA_LOG_WARN("Can not attach user now: transport was not set.");
-        }
+    UserAttachRequest *request = new UserAttachRequest;
+    request->userAccessToken = userAccessToken;
+    request->userExternalId = userExternalId;
+    request->userVerifierId = userVerifierToken;
+
+    if (listener != nullptr) {
```
izeigerman: if (listener) would be enough.

client/client-multi/client-cpp/impl/event/registration/EndpointRegistrationManager.cpp
```diff
@@ -284,25 +284,43 @@ void EndpointRegistrationManager::attachUser(const std::string& userExternalId
...
-        userAttachRequest_.reset(request);
+    KAA_LOG_INFO(boost::format("Going to attach to user '%1%' by access token: '%2%' (user verifier '%3%')")
+                                        % userExternalId % userAccessToken % DEFAULT_USER_VERIFIER_TOKEN);
 
-        if (userTransport_ != nullptr) {
-            userTransport_->sync();
-        } else {
-            KAA_LOG_WARN("Can not attach user now: transport was not set.");
-        }
+    UserAttachRequest *request = new UserAttachRequest;
```
ohamov: I'd substitute the raw new with unique_ptr to avoid memory leak

client/client-multi/client-cpp/test/impl/event/EndpointRegistrationManagerTest.cpp
```diff
@@ -71,14 +71,21 @@ BOOST_AUTO_TEST_CASE(UserAttachTest)
 
     TestEndpointAttachStatusListener* resultListener = new TestEndpointAttachStatusListener;
```
GamovCoder: if something below throws memory leaks

client/client-multi/client-cpp/kaa/event/registration/EndpointRegistrationManager.hpp
```diff
@@ -83,52 +81,40 @@ class EndpointRegistrationManager : public IEndpointRegistrationManager
...
+    KAA_MUTEX_DECLARE(attachEndpointGuard_);
+    KAA_MUTEX_DECLARE(detachEndpointGuard_);
 
-    UserTransport *                                            userTransport_;
+    std::unordered_map<std::int32_t, std::string/* endpoint access token */> attachEndpointRequests_;
+    std::unordered_map<std::int32_t, std::string/* endpoint key hash */>     detachEndpointRequests_;
 
-    KaaObservable<void (const AttachedEndpoints&), IAttachedEndpointListListener *> attachedEPListListeners_;
-    KAA_R_MUTEX_DECLARE(listenerGuard_);
+    std::unordered_map<std::int32_t, IAttachEndpointCallback*> attachEndpointListeners;
```
izeigerman: Add "_" to member's name.

client/client-multi/client-cpp/kaa/event/registration/EndpointRegistrationManager.hpp
```diff
@@ -83,52 +81,40 @@ class EndpointRegistrationManager : public IEndpointRegistrationManager
...
+    std::unordered_map<std::int32_t, IDetachEndpointCallback*> detachEndpointListeners;
```
izeigerman: Add "_" to member's name.

client/client-multi/client-cpp/kaa/event/registration/EndpointRegistrationManager.hpp
izeigerman: What if the user will add a new Endpoint Listener during the callback processing? The deadlock will occur. Why don't you use KaaObservable<void (const AttachedEndpoints&), std::int32_t> instead?

client/client-multi/client-cpp/kaa/event/registration/EndpointRegistrationManager.hpp
izeigerman: Same deadlock issue.

client/client-multi/client-cpp/kaa/event/registration/EndpointRegistrationManager.hpp
ghost: I agree, there is the possibility of the deadlock. But the IAttachEndpointCallback interface has two callback methods, so, as I understand, KaaObservable is not applicable to my case.

client/client-multi/client-cpp/impl/event/registration/EndpointRegistrationManager.cpp
```diff
@@ -18,331 +18,302 @@
...
-        }
-
-        userAttachRequest_.reset(request);
-
-        if (userTransport_ != nullptr) {
-            userTransport_->sync();
-        } else {
-            KAA_LOG_WARN("Can not attach user now: transport was not set.");
-        }
+    if (!strlen(DEFAULT_USER_VERIFIER_TOKEN)) {
```
izeigerman, line 266: Compile-time verification is better idea.

PR #130, opened 2015-03-05
KAA-402: Implement android demo application to demonstrate event API

ikulikov:


examples/photoframe/source/src/org/kaaproject/kaa/demo/photoframe/PhotoFrameController.java
```diff
@@ -0,0 +1,360 @@
...
+import org.kaaproject.kaa.demo.photoframe.event.StopPlayEvent;
+import org.kaaproject.kaa.demo.photoframe.event.UserAttachEvent;
+import org.kaaproject.kaa.demo.photoframe.event.UserDetachEvent;
+
+import android.content.Context;
+import android.database.Cursor;
+import android.net.Uri;
+import android.provider.MediaStore;
+import de.greenrobot.event.EventBus;
+
```
ashvayka, line 45: This class need some basic description/comments. Smth like:
This class receives Kaa events and user attach/detach callbacks and dispatch them to other application components via event bus. It is also responsible for fetching albums from android {@link MediaStore}.

PR #144, opened 2015-03-16
KAA-428: Kaa client outputs log warning on startup

batytskyy:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/AbstractLogCollector.java
```diff
@@ -91,8 +91,12 @@ public void setStorage(LogStorage storage) {
     public void fillSyncRequest(LogSyncRequest request) {
         LogBlock group = null;
         synchronized (storage) {
-            group = storage.getRecordBlock(strategy.getBatchSize());
             isUploading = false;
+            if (storage.getStatus().getConsumedVolume() == 0) {
+                LOG.debug("Log group is null: storage is empty");
```
ashvayka: please correct log message

PR #145, opened 2015-03-16
KAA-423: Collecting data from endpoints (C++ SDK demo app).

ghost:


examples/cpp-sdk-logging-sample/src/KaaDemo.cpp
```diff
@@ -0,0 +1,77 @@
...
+    logRecord.level = Level::DEBUG;
+    logRecord.tag = "Demo tag";
+    logRecord.message = "Demo log message";
+
+    /*
+     * Add the log record every second.
+     */
+    size_t logNumber = 0;
+    while (logNumber++ < LOG_NUMBER_TO_SEND) {
+        kaaClient.getLogCollector().addLogRecord(logRecord);
```
ohamov: I think I'd be better if we just once get LogCollector instead of doing it inside the circle

izeigerman:
LGTM

PR #149, opened 2015-03-18
Kaa-126: Improve Java SDK thread model

ashvayka:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/AbstractKaaClient.java
```diff
@@ -233,32 +230,30 @@
...
         logTransport.setLogProcessor(logCollector);
+    }
 
-        for (KaaTransport transport : transports.values()) {
-            transport.setChannelManager(channelManager);
-            transport.setClientState(kaaClientState);
-        }
-
-        channelManager.setConnectivityChecker(context.createConnectivityChecker());
+    private void initTransport(KaaTransport transport) {
```
shpigor: Maybe it better to add initTransport(List<KaaTransport> transport) method with argument Arrays.asList({bootstrapTransport, profileTransport, ...})

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/impl/DefaultChannelManager.java
```diff
@@ -49,6 +53,9 @@
     private final Map<TransportProtocolId, List<TransportConnectionInfo>> bootststrapServers;
     private final Map<TransportProtocolId, TransportConnectionInfo> lastBSServers = new HashMap<>();
 
+    private final Map<String, BlockingQueue<SyncTask>> syncTaskQueueMap = new HashMap<String, BlockingQueue<SyncTask>>();
```
shpigor: Maybe better to use ConcurrencyHashMap

PR #155, opened 2015-03-23
KAA-429: Using notifications (C++ SDK demo app).

ghost:


client/client-multi/client-cpp/kaa/observer/KaaObservable.hpp
```diff
@@ -90,6 +90,13 @@ class KaaObservable
         slotsToRemove_.clear();
     }
 
+    bool isEmpty()
+    {
+        KAA_MUTEX_UNIQUE_DECLARE(lock, mainGuard_);
+        KAA_MUTEX_UNIQUE_DECLARE(modLock, modificationGuard_);
+        return slots_.empty() && slotsToAdd_.empty();
```
yzeygerman: I guess OR condition is a better idea.

PR #157, opened 2015-03-24
KAA-441: Sample application to demonstrate C++ Event subsystem

izeigerman:
Reviewers @dyosick  and @GamovCoder


client/client-multi/client-cpp/kaa/common/AvroByteArrayConverter.hpp
```diff
@@ -120,10 +120,6 @@ AvroByteArrayConverter<T>::AvroByteArrayConverter()
 template<typename T>
 T AvroByteArrayConverter<T>::fromByteArray(const std::uint8_t* data, const std::uint32_t& dataSize)
 {
-    if (!data || dataSize == 0) {
```
ghost, line 123: I think we should leave this check, but change the comparison operation to XOR.

client/client-multi/client-cpp/kaa/common/AvroByteArrayConverter.hpp
```diff
@@ -136,10 +132,6 @@ T AvroByteArrayConverter<T>::fromByteArray(const std::uint8_t* data, const std::
 template<typename T>
 void AvroByteArrayConverter<T>::fromByteArray(const std::uint8_t* data, const std::uint32_t& dataSize, T& datum)
 {
-    if (!data || dataSize == 0) {
```
ghost, line 139: The same as above.

server/control/src/main/java/org/kaaproject/kaa/server/control/service/sdk/event/CppEventSourcesGenerator.java
```diff
@@ -163,10 +170,17 @@
                     if (eventFamilyListenersOnGenericEvent.length()>0) {
                         eventFamilyListenersOnGenericEvent += "else ";
                     }
+                    if (eventFamilyListenersNotifyListener.length()>0) {
```
ohamov: please add spaces between eventFamilyListenersNotifyListener.length() and ">" & ">" and "0" respectively

PR #162, opened 2015-03-25
KAA-448: Need to disable ability to edit existing topic.

shpigor:


ikulikov:
TopicView  should be read only when create == false. Save button should be hidden.

PR #163, opened 2015-03-25
KAA-442:Can't generate complex configuration schema with recursion

izeigerman:
Reviewers @ashvayka and @Shpigor


shpigor:
Reviewed

PR #164, opened 2015-03-25
KAA-35: Overdue expiry time for notification

serye:


server/admin/src/main/java/org/kaaproject/kaa/server/admin/services/KaaAdminServiceImpl.java
```diff
@@ -1940,6 +1940,9 @@ public EndpointNotificationDto sendUnicastNotification(
             throws KaaAdminServiceException {
         checkAuthority(KaaAuthorityDto.TENANT_DEVELOPER, KaaAuthorityDto.TENANT_USER);
         try {
+            if(null!=notification.getExpiredAt() && notification.getExpiredAt().before(new Date())){
```
ashvayka: Usually we write notification.getExpiredAt() != null. + spaces.

PR #191, opened 2015-04-02
KAA-461: C SDK a lot of stack is used

ikulikov:


client/client-multi/client-c/src/kaa/kaa_event.c
```diff
@@ -706,30 +714,45 @@ static kaa_error_t kaa_event_read_event(kaa_event_manager_t *self, kaa_platform_
...
         if (error) {
             KAA_LOG_ERROR(self->logger, error, "Failed to read event data size field");
-            return error;
+            goto end;
         }
         event_data_size = KAA_NTOHL(event_data_size);
     }
 
-    kaa_endpoint_id event_source;
+    event_source = (uint8_t *)KAA_MALLOC(KAA_ENDPOINT_ID_LENGTH);
```
ghost: I doubt that there is a strong need to allocate this buffer every time.
I suggest to create the buffer once either in a static manner or maybe as the member of the event manager.

client/client-multi/client-c/src/kaa/kaa_event.c
```diff
@@ -681,15 +681,23 @@ kaa_error_t kaa_event_request_serialize(kaa_event_manager_t *self, size_t reques
 
 static kaa_error_t kaa_event_read_event(kaa_event_manager_t *self, kaa_platform_message_reader_t *reader)
 {
-    KAA_RETURN_IF_NIL2(self, reader, KAA_ERR_BADPARAM);
+
+    kaa_error_t error = KAA_ERR_NONE;
+    char *event_fqn = NULL;
+    uint8_t* event_source = NULL;
+
+    if (!(self) || !(reader)) {
```
ghost: Please, remove enclosing parentheses all around this function.

client/client-multi/client-c/src/kaa/kaa_event.c
```diff
@@ -681,15 +681,23 @@ kaa_error_t kaa_event_request_serialize(kaa_event_manager_t *self, size_t reques
...
+        error = KAA_ERR_BADPARAM;
+        goto end;
```
ghost: According to the approved code style we don't use goto.
I know, your suggestion is good, but we have the rule.

client/client-multi/client-c/src/kaa/kaa_profile.c
```diff
@@ -240,6 +240,7 @@ static kaa_error_t kaa_version_info_serialize(kaa_platform_message_writer_t *wri
         error_code = kaa_platform_message_write(writer, &field_number_with_reserved, sizeof(uint16_t));
         KAA_RETURN_IF_ERR(error_code);
 
+        network_order_schema_version = 0;
```
ghost: Could you retest without this "fix"? Is the problem is still reproduced?
If it does, please, write the explanation comment why we really need this assignment.

PR #197, opened 2015-04-16
KAA-492: Improve log uploading to work on schedule, not only when a record is added

batytskyy:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/AbstractLogCollector.java
```diff
@@ -231,4 +227,15 @@ public void run() {
             }, delay, TimeUnit.SECONDS);
         }
     }
+
+    public final void scheduleAtFixedRateLogUpload() {
```
ashvayka: This is wrong implementation. We do not need to schedule some periodic checks in all cases. We need to schedule only one check when log is actually sent.

batytskyy:
These changes have been used in bug/KAA-537 branch.

PR #200, opened 2015-04-21
KAA:494: Process REFUSE_BAD_CREDENTIALS from server on Java client

batytskyy:


server/operations/src/main/java/org/kaaproject/kaa/server/operations/service/profile/DefaultProfileService.java
```diff
@@ -121,6 +126,8 @@ public EndpointProfileDto registerProfile(RegisterProfileRequest request) {
             dto.setNfSequenceNumber(0);
             dto.setChangedFlag(Boolean.FALSE);
 
+            cacheService.putEndpointKey(keyHash, generateEndpointKey(dto.getEndpointKey()));
```
ashvayka: Use KeyUtil.getPublic() here instead of your own method

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/impl/channels/DefaultOperationTcpChannel.java
```diff
@@ -116,8 +116,13 @@ public void run() {
         @Override
         public void onMessage(ConnAck message) {
             LOG.info("ConnAck ({}) message received for channel [{}]", message.getReturnCode(), getId());
+
             if (message.getReturnCode() != ReturnCode.ACCEPTED) {
                 LOG.error("Connection for channel [{}] was rejected: {}", getId(), message.getReturnCode());
+                if (message.getReturnCode() == ReturnCode.REFUSE_BAD_CREDENTIALS) {
```
ashvayka, line 122: This type of error is not specific to TCP channel, and should also be handled in HTTP channel

PR #208, opened 2015-05-12
KAA-500: Optimize memory usage of Kaa sandbox

batytskyy:


ashvayka:
This feature should be re-implemented. You need to analyze node max memory and base on this run-time info enable different java options

PR #210, opened 2015-05-14
KAA-502: Add Cassandra to sandbox

batytskyy:


sandbox/builder/src/main/java/org/kaaproject/kaa/sandbox/AbstractSandboxBuilder.java
```diff
@@ -344,6 +344,10 @@ protected void initBoxData() throws Exception {
             }
         }
 
+        LOG.info("Executing Cassandra cql script...");
+        executeSudoSsh("cqlsh -f "+CASSANDRA_INIT_SCRIPT);
```
ashvayka, line 349: spaces (formatting)

sandbox/builder/src/main/java/org/kaaproject/kaa/sandbox/AbstractSandboxBuilder.java
```diff
@@ -344,6 +344,10 @@ protected void initBoxData() throws Exception {
...
+        Thread.sleep(5000L);
```
ashvayka: Why do we need this?

ashvayka:
Some minor questions, otherwise, LGTM

sandbox/builder/src/main/java/org/kaaproject/kaa/sandbox/AbstractSandboxBuilder.java
```diff
@@ -344,6 +344,10 @@ protected void initBoxData() throws Exception {
             }
         }
 
+        LOG.info("Executing Cassandra cql script...");
```
ashvayka, line 348: We need to control AbstractSandboxBuilder behavior using arguments. In other words, we need a dedicated argument that will trigger either Cassandra setup script or Mongo DB or both (all).

sandbox/builder/src/main/java/org/kaaproject/kaa/sandbox/AbstractSandboxBuilder.java
batytskyy: It is approximate time of Cassandra init script execution.

PR #211, opened 2015-05-14
KAA-427: Java EP SDK contains thirdparty libraries

batytskyy:


ashvayka:
This changes are ok, however, lets merge them after iot_world branch, in order to make sure that all iot_world related applications also work.

PR #212, opened 2015-05-14
KAA-503: dump sandbox build logs on fail

serye:


sandbox/builder/src/main/java/org/kaaproject/kaa/sandbox/AbstractSandboxBuilder.java
```diff
@@ -523,6 +525,22 @@ private void transferFile(String file, String to) {
...
+        File dumpedLogsFolder  = new File(LOG_DUMP_LOCATION);
+        if(!dumpedLogsFolder.exists()){
+            dumpedLogsFolder.mkdirs();
+        }
+        Scp scp = createScp();
+        scp.setLocalTodir(LOG_DUMP_LOCATION);
+        scp.setFile(SSH_USERNAME+"@"+sshForwardPort+":/var/log/kaa/*.log");
+        scp.execute();
+        try {
+            Thread.sleep(2);
```
ashvayka: Why do you need this?

serye: Error code 500 with no reason of fail. Every time I wanna know what is the cause of build fail I should comment some lines in AbstractSanboxBuilder and recompile whole module just to read logs. It takes a lot of time and really annoying.

PR #236, opened 2015-06-10
cell_monitor_logback: use default Android logger instead of logback

batytskyy:
Logback was eliminated from client SDK, so there is no need to add one more dependency to the final application.


batytskyy:
Logback isn't used, SL4J Android wraps default logger.

PR #238, opened 2015-06-10
KAA-515: Persistent log data storage for Java/Android EP SDK

batytskyy:


client/client-multi/client-java-desktop/src/main/java/org/kaaproject/kaa/client/logging/DesktopSQLiteDBLogStorage.java
```diff
@@ -0,0 +1,379 @@
...
+    @Override
+    public LogBlock getRecordBlock(long blockSize) {
+        synchronized (connection) {
+            LOG.trace("Creating a new record block, needed size: {}", blockSize);
+            Statement statement = null;
+            ResultSet rs = null;
+            LogBlock logBlock = null;
+            List<String> unmarkedRecordIds = new LinkedList<>();
+            List<LogRecord> logRecords = new LinkedList<>();
+            try {
```
serye, line 121: maybe it would be better to use java7 try-catch-with-resources syntax? 
        try(Statement statement = connection.createStatement();
            ResultSet resultSet = statement.executeQuery("")){  
        }

and there is no need in methods like tryCloseResultSet(rs), and finally block at all

client/client-multi/client-java-desktop/src/main/java/org/kaaproject/kaa/client/logging/DesktopSQLiteDBLogStorage.java
```diff
@@ -0,0 +1,379 @@
...
+                    throw new RuntimeException("Unable to retrieve consumed size and volume");
+                }
+            } finally {
+                tryCloseResultSet(rs);
+                tryCloseStatement(statement);
+            }
+        }
+    }
+
+    public void close() {
```
serye, line 336: give more meaningful name to this method, it isn`t obviously what does it close

batytskyy, line 336: Method's explanation is provided in LogStorage interface. Please propose more suitable name.

PR #249, opened 2015-06-17
KAA-127: Improve C++ SDK thread model

ghost:


client/client-multi/client-cpp/impl/context/SimpleExecutorContext.cpp
```diff
@@ -0,0 +1,49 @@
...
+    if (!lifeCycleThreadCount_ || !apiThreadCount_ || !callbackThreadCount_) {
+        KAA_LOG_ERROR(boost::format("Failed to create executor context: lifeCycleThreadCount %u, apiThreadCount_ %u, callbackThreadCount_ %u,")
+            % lifeCycleThreadCount_ % apiThreadCount_ % callbackThreadCount_);
+        throw KaaException("Failed to crate executor context: bad input parameters");
+    }
+}
+
+void SimpleExecutorContext::init()
+{
+    lifeCycleExecutor_ = createExecutor(DEFAULT_THREAD_COUNT);
```
ohamov: I consider we should use lifeCycleThreadCount_, apiThreadCount_, lifeCycleThreadCount_ instead of default parameters

client/client-multi/client-cpp/kaa/utils/ThreadPool.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include "kaa/utils/KaaTimer.hpp"
+#include "kaa/utils/IThreadPool.hpp"
+
+namespace kaa {
+
+class ThreadPool : public IThreadPool {
+    friend class Worker;
+
+public:
+    ThreadPool() : ThreadPool(DEFAULT_WORKER_NUMBER) {}
```
GamovCoder: Do we really need to use delegating constructor? I'd not rely on compiler to reduce unnecessary function call.Why don't you use parameters by default?

client/client-multi/client-cpp/impl/context/SimpleExecutorContext.cpp
ghost: Thanks a lot, it's my fault.

client/client-multi/client-cpp/kaa/context/SimpleExecutorContext.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+#include "kaa/context/AbstractExecutorContext.hpp"
+
+#include <memory>
+
+namespace kaa {
+
+class SimpleExecutorContext : public AbstractExecutorContext {
+public:
+    SimpleExecutorContext() : SimpleExecutorContext(DEFAULT_THREAD_COUNT, DEFAULT_THREAD_COUNT, DEFAULT_THREAD_COUNT) {}
```
ohamov: Can we use parameters by default instead of delegating constructor ?

ghost: Ok

PR #257, opened 2015-06-22
Kaa 525/events

ohamov:


client/client-multi/client-c/src/kaa/kaa_event.c
```diff
@@ -586,7 +588,8 @@ static kaa_error_t kaa_event_list_serialize(kaa_event_manager_t *self, kaa_list_
                 return error;
             }
         }
-
+        KAA_LOG_TRACE(self->logger, KAA_ERR_NONE, "Serialized event: sqn '%u',fqn length '%u', data size '%u'"
```
ghost: Please, print out fqn instead of its length

PR #297, opened 2015-08-27
Kaa-318: Improve performance of generic flume log appender.

obantysh:
Used thread pool and RPC connections pool for sending logs, instead of single RPC client implementation. This mechanism also fixed KAA-639 bug


server/appenders/flume/appender/src/main/java/org/kaaproject/kaa/server/appenders/flume/appender/FlumeLogAppender.java
```diff
@@ -30,41 +37,90 @@
...
-                        LOG.warn("Flume client wasn't initialized. Invoke method init before.");
-                        listener.onInternalError();
-                    }
-                }
-                else {
-                    LOG.warn("Unable to generate Flume events from log event pack!");
+            if (executor == null || callbackExecutor == null || flumeClientManager == null) {
+                try {
+                    reinit();
+                } catch (NullPointerException e) {
```
ashvayka: Who is throwing NPE and why? Is this thirdparty (flume) code?

PR #299, opened 2015-08-28
KAA-582: Add 2 empty value checks

krupet:


server/admin/src/main/java/org/kaaproject/kaa/server/admin/client/mvp/view/widget/RecordPanel.java
```diff
@@ -162,9 +162,14 @@ public void reset() {
...
+        return recordFieldWidget.validate() && isEmpty(recordFieldWidget);
     }
-    
+
+    private boolean isEmpty(RecordFieldWidget recordFieldWidget) {
+        return !(recordFieldWidget.getValue() != null && !recordFieldWidget.getValue().getValue().isEmpty());
+    }
+
+
     public void setFormDataLoader(FormDataLoader formDataLoader) {
```
ikulikov, line 169: Don't use isEmpty in validate method. isEmpty should be separate public method. Also recordFieldWidget is already member of RecordPanel class, just access it directly without passing it as argument.

server/admin/src/main/java/org/kaaproject/kaa/server/admin/client/mvp/view/event/EcfSchemaViewImpl.java
```diff
@@ -125,12 +125,16 @@ protected void resetImpl() {
...
             return true;
         }
     }
-    
+
+    private boolean isEmpty(RecordPanel ecfSchemaForm) {
+        return !(ecfSchemaForm.getValue() != null && !ecfSchemaForm.getValue().getValue().isEmpty());
+    }
+
     @Override
```
ikulikov, line 151: Use public method isEmpty of RecordPanel class, see comment above.

server/admin/src/main/java/org/kaaproject/kaa/server/admin/client/mvp/view/event/EcfSchemaViewImpl.java
```diff
@@ -125,12 +125,16 @@ protected void resetImpl() {
...
     protected boolean validate() {
         if (create) {
-            return ecfSchemaForm.validate();
+            return ecfSchemaForm.validate() && isEmpty(ecfSchemaForm);
         } else {
```
ikulikov, line 146: Schema form shouldn't be empty, add '!'.

PR #301, opened 2015-08-31
KAA-646: Add message if default user verifier is not specified with Ecf

krupet:


server/admin/src/main/java/org/kaaproject/kaa/server/admin/client/mvp/activity/GenerateSdkActivity.java
```diff
@@ -175,19 +175,25 @@ protected void getEntity(String id, AsyncCallback<SdkPropertiesDto> callback) {}
...
         onSave();
-        KaaAdmin.getDataSource().generateSdk(entity,new BusyAsyncCallback<String>() {
 
-            @Override
-            public void onFailureImpl(Throwable caught) {
-                Utils.handleException(caught, detailsView);
-            }
+        if (!entity.getAefMapIds().isEmpty() && entity.getDefaultVerifierToken() == null) {
+            detailsView.setErrorMessage("Specify default user verifier.");
+        } else {
```
ikulikov, line 182: Please use i18n lookup constants for text message. See https://github.com/kaaproject/kaa/blob/master/server/admin/src/main/resources/org/kaaproject/kaa/server/admin/client/i18n/KaaAdminConstants.properties

PR #303, opened 2015-09-04
KAA-652: implements thread save method for getting configration in Kaa Java client api

sergeyevd:


server/control/src/main/resources/sdk/java/cf/ConfigurationDeserializer.java.template
```diff
@@ -45,6 +45,8 @@ class ConfigurationDeserializer {
     }
 
     ${configuration_class} fromByteArray(byte[] data) throws IOException {
-        return converter.fromByteArray(data);
+        synchronized(converter){
```
ashvayka: bad formatting

PR #304, opened 2015-09-07
Kaa-645: Review and update REST API documentation

ashvayka:


server/admin/src/main/java/org/kaaproject/kaa/server/admin/controller/KaaAdminController.java
```diff
@@ -950,11 +958,29 @@ public void removeTopicFromEndpointGroup(
     @ResponseBody
     public NotificationDto sendNotification(
             @RequestPart("notification") NotificationDto notification,
+            @RequestPart("ttl") String ttl,
```
ashvayka: Let's remove this, because we have expireAt in NotificationDto

server/admin/src/main/java/org/kaaproject/kaa/server/admin/services/KaaAdminServiceImpl.java
```diff
@@ -230,11 +226,29 @@ public TenantUserDto getTenant(String userId) throws KaaAdminServiceException {
         }
     }
 
+    private void checkTenantName(TenantUserDto tenantUser) throws KaaAdminServiceException {
+        String name = tenantUser.getTenantName();
+        if (isEmpty(name)) {
+            throw new IllegalArgumentException("Parameter tenantName is empty.");
+        }
+        List<TenantUserDto> tenants = new ArrayList<>();
+        tenants.addAll(getTenants());
```
ashvayka: This is redundant. You should do this chek inside saveUser, or add dedicated query for this. We can't fetch all tenants each time we need to do validation

server/admin/src/main/java/org/kaaproject/kaa/server/admin/services/util/Utils.java
```diff
@@ -51,8 +51,9 @@ public static KaaAdminServiceException handleExceptionWithCause(Exception except
 
         if (exceptionWithCause instanceof KaaAdminServiceException) {
             return (KaaAdminServiceException) exceptionWithCause;
-        } else if (exceptionWithCause instanceof IllegalArgumentException) {
-        	return new KaaAdminServiceException(exceptionWithCause.getMessage(), ServiceErrorCode.BAD_REQUEST_PARAMS);
+        } else if (exceptionWithCause instanceof IllegalArgumentException || 
+                cause.equals("org.kaaproject.kaa.server.common.dao.exception.IncorrectParameterException")) {
```
ashvayka: please replace this with cause.contains("IncorrectParameterException")

PR #309, opened 2015-09-10
KAA-651: check sdk token value before reading configuration data

batytskyy:


ashvayka:
I believe much better solution is
1) store sdkToken in state file.
2) check that sdkToken was changed on start of the client.
3) if sdk token was changed - cleanup stored configuration and it's hash.
4) send profile update to invalidate endpoint group list, etc.
5) persist new sdkToken to state file only when you have received update from server.

PR #336, opened 2015-10-15
KAA-701: Control server should notify operations server when endpoint group is deleted

ashvayka:


ashvayka:
@Shpigor @ikulikov please review!

ikulikov:
LGTM

PR #342, opened 2015-10-22
KAA-625: [Java] SDK should notify if custom profile is not set.

ghost:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/AbstractKaaClient.java
```diff
@@ -205,12 +205,14 @@
...
         transportContext.getLogTransport().setLogProcessor(logCollector);
         transportContext.initTransports(this.channelManager, this.kaaClientState);
-        
+
         eventFamilyFactory = new EventFamilyFactory(eventManager, context.getExecutorContext());
     }
 
     @Override
-    public void start() {
+    public void start() throws KaaException {
```
ashvayka: Instead of throwing exception here, it is better to execute

```
                    if (stateListener != null) {
                        stateListener.onStartFailure(new KaaException("your text"));
                    }
```

ghost: I suggest to use both ways. If a listener is specified, notify it, otherwise, throw an exception.

ashvayka:
LGTM

PR #346, opened 2015-10-22
KAA-570: [Java SDK] Add warning messages as a result of API calls when client is in invalid state.

ghost:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/AbstractKaaClient.java
```diff
@@ -205,12 +230,15 @@
...
         transportContext.initTransports(this.channelManager, this.kaaClientState);
-        
+
         eventFamilyFactory = new EventFamilyFactory(eventManager, context.getExecutorContext());
     }
 
     @Override
-    public void start() {
+    public void start() throws KaaException {
+        checkClientStateNot(State.STARTED, "Kaa client is already run");
```
ashvayka: is already started?

batytskyy:
Do we really need to change Kaa Client SDK API declaring that KaaClient's methods throw KaaException?

ashvayka:
LGTM

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/AbstractKaaClient.java
```diff
@@ -155,6 +155,31 @@
...
     protected final KaaClientStateListener stateListener;
 
+    protected enum State {
+        CREATED,
+        STARTED,
+        PAUSED,
+        STOPPED
+    };
+
+    protected State clientState = State.CREATED;
```
shpigor, line 167: Maybe better to move to the end of constructor

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/AbstractKaaClient.java
```diff
@@ -375,46 +420,51 @@ public void removeNotificationListener(String topicId, NotificationListener list
 
     @Override
     public void subscribeToTopic(String topicId) throws UnavailableTopicException {
-        this.notificationManager.subscribeToTopic(topicId, true);
+        subscribeToTopic(topicId, true);
```
shpigor: I think better to use constants FORCE_SYNC and WITHOUT_SYNC instead bool value.

shpigor:
Revieved

PR #349, opened 2015-10-28
KAA-126: Improve Java SDK thread model

batytskyy:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/configuration/base/AbstractConfigurationManager.java
```diff
@@ -89,7 +92,16 @@ public void processConfigurationData(ByteBuffer buffer, boolean fullResync) thro
                         storage.saveConfiguration(ByteBuffer.wrap(configurationData));
                         LOG.debug("Persisted configuration data from storage {}", storage);
                     }
-                    deserializer.notify(Collections.unmodifiableCollection(listeners), configurationData);
+                    executorContext.getCallbackExecutor().submit(new Runnable() {
+                        @Override
+                        public void run() {
+                            try {
+                                deserializer.notify(Collections.unmodifiableCollection(listeners), configurationData);
```
ashvayka: We need to pick a strategy and decide either to notify each listener in separate task or notify all listeners in one task.
Anyway, if one listener will fail to process configuration update, other listeners should still receive this update. Same logic should be present for all listeners in our client code.

PR #351, opened 2015-10-29
Kaa-521: Endpoint profile access REST API

sashadidukh:
Added possibility to get profile based on endpoint key and on endpoint group id from cassandra and mongodb via REST API. Provided pagination feature when got profile by endpoint group id.


ashvayka:
This is already merged as 1 commit

PR #360, opened 2015-11-13
KAA-737: Add support of custom extensions in Avro platform protocol.

ghost:


common/endpoint-shared/src/main/avro/endpoint.avsc
```diff
@@ -434,6 +434,15 @@
  ]
 },
 {
+    "namespace": "org.kaaproject.kaa.common.endpoint.gen",
+    "type": "record",
+    "name": "PluginSync",
+    "fields": [
+        {"name": "pluginId", "type": "int"},
```
ashvayka: let's rename this to "extensionId" or "extId" in order to match design and avoid confusion. This is actually id of plugin instance in scope of specific sdk profile.

PR #367, opened 2015-11-23
KAA-606: CTL schema export

bkhablenko:


ashvayka:
Please extract CTLSchemaParser from KaaAdminServiceImpl to separate file

PR #388, opened 2015-12-08
KAA-761: Remove deprecated API to create singleton instance of C++ SDK.

ghost:


client/client-multi/client-cpp/impl/channel/connectivity/PingConnectivityChecker.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+#include <boost/asio/ip/tcp.hpp>
+#include <boost/asio/connect.hpp>
+#include <boost/asio/io_service.hpp>
+
+#include "kaa/logging/Log.hpp"
+
+namespace kaa {
+
+const std::string PingConnectivityChecker::DEFAULT_HOST = "www.google.com";
```
Karnil, line 31: why are you using "google.com" as the default host? what would happen if that host is not available?

ghost, line 31: It was made only for compatibility with Java SDK.
Insied C++ SDK uses its own connectivity checker than pings one of Kaa Bootstrap server.
See details in https://github.com/kaaproject/kaa/blob/master/client/client-multi/client-cpp/impl/channel/connectivity/IPConnectivityChecker.cpp

client/client-multi/client-cpp/kaa/KaaThread.hpp
```diff
@@ -19,7 +19,7 @@
 
 #include "kaa/KaaDefaults.hpp"
 
-#ifdef KAA_THREADSAFE
+//#ifdef KAA_THREADSAFE
```
Karnil, line 22: Maybe it should be removed?

client/client-multi/client-cpp/kaa/KaaThread.hpp
```diff
@@ -57,7 +57,7 @@ typedef std::atomic_int_fast32_t RequestId;
 
 #define kaa_thread_local thread_local
 
-#else
+#if 0
```
Karnil, line 60: You should to remove this block.

ghost, line 60: I agree, but let me do this in other task.

client/client-multi/client-cpp/kaa/KaaThread.hpp
ghost, line 22: I agree, but let me do this in other task.

PR #391, opened 2015-12-11
Remove deque

rasendubi:
- It's not used at all
- Deque is the same thing as double-linked list. Both underlying
  structure and implementation are the same modulo alpha-conversion.


ghost:
LGTM

Karnil:
LGTM

PR #413, opened 2015-12-23
Feature/kaa 764

obantysh:


rasendubi:
v0.8.0 is already released. If changes are still necessary, reopen against appropriate branch.

PR #415, opened 2015-12-24
KAA-780: [C SDK] Make the amount of parallel log upload batches configurable.

ghost:


client/client-multi/client-c/src/kaa/kaa_common_schema.h
```diff
@@ -38,7 +38,7 @@ extern "C" {
 
 
 typedef void (*serialize_fn)(avro_writer_t writer, void *data);
-typedef void *(*deserialize_fn)(void);
+typedef void *(*deserialize_fn)(); // NOTE: Do not add void in the parameter list.
```
rasendubi: Shouldn't the function take `void *`?

client/client-multi/client-c/test/test_kaa_log.c
```diff
@@ -69,6 +67,15 @@ static kaa_logger_t *logger = NULL;
 static kaa_status_t *status = NULL;
 static kaa_channel_manager_t *channel_manager = NULL;
 
+static kaa_service_t OPERATIONS_SERVICES[] = { KAA_SERVICE_PROFILE
```
rasendubi: Shouldn't this be `const kaa_service_t`?

client/client-multi/client-c/src/kaa/kaa_common_schema.h
ghost: deserialize_fn is a something like a parent class for deserialize_wo_ctx_fn and deserialize_w_ctx_fn. The signature of these typedefs of pointer to functions is different so we can't use void*

Maybe it isn't a good practice to use this kind of hack, but now to make it to more clear a lot of code need to be re-implemented in Avro Gen C.

client/client-multi/client-c/test/test_kaa_log.c
ghost: Sure, I will do.

client/client-multi/client-c/src/kaa/kaa_common_schema.h
rasendubi: Could you put that in a comment or open a ticket, please?

ghost: Added a comment.
I suggest to discuss improvements of Avro C and we will open a story. Ok?

PR #426, opened 2016-01-06
[Java SDK] Add callback/future for log delivery

vpogorelovmapr:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/AbstractLogCollector.java
```diff
@@ -147,8 +147,20 @@ public void onLogResponse(LogSyncResponse logSyncResponse) throws IOException {
...
+                bucketInfoMap.remove(requestId);
                 if (response.getResult() == SyncResponseResultType.SUCCESS) {
                     storage.removeRecordBlock(response.getRequestId());
+                    executorContext.getCallbackExecutor().execute(new Runnable() {
+                        @Override
+                        public void run() {
+                            if (logDeliveryListener != null) {
+                                logDeliveryListener.onLogDeliverySuccess(bucketInfo);
+                            }
+                            futureMap.get(requestId).setValue(bucketInfo);
```
dyosick: Why the value of a future is set in a executor task?

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/future/BucketFuture.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public boolean cancel(boolean mayInterruptIfRunning) {
+        throw new RuntimeException("Not implemented");
+    }
+
+    @Override
+    public boolean isCancelled() {
+        return false;
```
ghost: Maybe "throw new RuntimeException("Not implemented");" is better?

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/future/BucketFuture.java
```diff
@@ -0,0 +1,61 @@
+package org.kaaproject.kaa.client.logging.future;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.*;
+
+public class BucketFuture<T> implements Future<T> {
```
ghost: As this feature is for 0.8.0 release, it means what Java 8 may be used.
So I suggest to use https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/PersistentLogStorageConstants.java
```diff
@@ -19,59 +19,89 @@
...
+    String KAA_UPDATE_STORAGE_INFO =
+                    "INSERT OR REPLACE INTO  " + STORAGE_INFO_TABLE_NAME + " " +
+                    "(" + STORAGE_INFO_KEY_COLUMN + "," + STORAGE_INFO_VALUE_COLUMN + ") " +
+                    "VALUES (?, ?);";
+
+    String KAA_SELECT_STORAGE_INFO =
+                    "SELECT " + STORAGE_INFO_VALUE_COLUMN + " " +
+                    "FROM " + STORAGE_INFO_TABLE_NAME + " " +
+                    "WHERE " + STORAGE_INFO_KEY_COLUMN + " = ?;";;
+
```
ghost, line 106: Double ';' at the end)))

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/future/ExecutionResult.java
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kaaproject.kaa.client.logging.future;
+
+public class ExecutionResult<T> {
```
ghost, line 18: Please, see the comment above about CompletableFuture

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/future/BucketFuture.java
ashvayka: No, this method is important. You can not cancel our future, thus isCancelled should always return false. It is used in executors.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/future/BucketFuture.java
ashvayka: It may cause issues on Android. http://stackoverflow.com/questions/23318109/is-it-possible-to-use-java-8-for-android-development

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/KaaClient.java
```diff
@@ -55,7 +56,7 @@
      * @param record
      *            New log record object
      */
-    void addLogRecord(Log record);
+    BucketFuture addLogRecord(Log record);
```
ashvayka: This is wrong. as far as i understand, BucketFuture is generic class. We should return Future<BucketInfo> here.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/LogCollector.java
```diff
@@ -44,6 +45,10 @@
      * Adds new log record to local storage.
      *
      * @param record New log record object
+     *
+     * @return bucketinfo object for added record
      */
-    void addLogRecord(Log record);
+    BucketFuture<BucketInfo> addLogRecord(Log record);
+
+    void setLogDeliveryListener(LogDeliveryListener listener);
```
ashvayka: Is this method available to user? I don't see it in Kaa client

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/memory/MemLogStorage.java
```diff
@@ -101,15 +97,12 @@ public void addLogRecord(LogRecord record) {
...
     }
 
     @Override
-    public LogBlock getRecordBlock(long blockSize, int batchCount) {
-        LOG.trace("Getting new record block with block size = {} and count = {}", blockSize, batchCount);
-        if (blockSize > maxBucketSize || batchCount > maxBucketRecordCount) {
-            //TODO: add support of block resize
-            LOG.warn("Resize of record block is not supported yet");
-        }
+    public LogBlock getRecordBlock() {
```
ashvayka: I think this method should be renamed/changed to 
public Bucket getNextBucket();
This will help to avoid confusuions betweeb BucketInfo and LogBlock. LogBlock is actually a Bucket. correct?

client/client-multi/client-java-android/src/templates/log/DefaultLogCollector.java.template
```diff
@@ -38,12 +41,15 @@ public class DefaultLogCollector extends AbstractLogCollector {
     }
 
     @Override
-    public void addLogRecord(final ${log_record_class_package}.${log_record_class} record) {
+    public BucketFuture<BucketInfo> addLogRecord(final ${log_record_class_package}.${log_record_class} record) {
+        final BucketFuture<BucketInfo> future = new BucketFuture<>();
```
ghost: By design, one future relates to one log bucket. According to this code snippet, a new future will be created for every log record. In the result, almost all future will never be completed.
Please, correct me if I am wrong.

ashvayka: Yes, this is wrong.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/memory/MemLogStorage.java
ghost: Correct. I will rename this method and corresponding classes.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/logging/LogCollector.java
ghost: Yes, it is available though the GenericKaaClient interface.
Also I will move this method to GenericLogCollector, because it doesn't require any auto-generated information.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/KaaClient.java
ghost: So should a raw Future object be returned or a template parameter should be specified as well?
Do we need to apply these changes to LogCollector.java?

PR #428, opened 2016-01-13
KAA-807: Code refactoring according to sonarqube rules.

sashadidukh:


sashadidukh:
We need to merge into 'develop/v0.8.1'.

PR #438, opened 2016-01-18
KAA-816: Introduce hasUpdate flag in client state to indicate if the state should be persisted. Add call to persist client state after each server response processing.

ikulikov:


ghost:
LGTM

ashvayka:
LGTM

PR #439, opened 2016-01-19
Kaa 773 log delivery callbacks

forGGe:
Includes:
- implementation
- tests
- test framework updates and small nitpicks


forGGe:
Another pull request will be provided

PR #451, opened 2016-01-22
Kaa 824 calculate hash from list of topics

aburdeinyi:
http://jira.kaaproject.org/browse/KAA-824


client/client-multi/client-c/src/kaa/collections/kaa_list.c
```diff
@@ -309,3 +309,17 @@ void kaa_list_sort(kaa_list_t *list, match_predicate pred)
     }
     list->tail = node;
 }
+
+uint32_t kaa_list_hash(kaa_list_t *list, list_node_hash pred)
```
ghost: Just interesting, would it be a problem what other SDKs use signed types - int and long?

client/client-multi/client-c/src/kaa/collections/kaa_list.c
```diff
@@ -309,3 +309,17 @@ void kaa_list_sort(kaa_list_t *list, match_predicate pred)
...
+
+int32_t kaa_list_hash(kaa_list_t *list, list_node_hash pred)
+{
+    KAA_RETURN_IF_NIL2(list, pred, 0);
+    int32_t result = 1;
+    kaa_list_node_t *node = kaa_list_begin(list);
+    uint64_t element;
+    while (node) {
+        element = pred(node->data);
+        result = 31 * result + (uint32_t) (element ^ (element >> 32));
```
ohamov, line 321: I'd avoid magic numbers, could you please substitute them with macro?

client/client-multi/client-c/src/kaa/kaa_notification_manager.c
```diff
@@ -1026,6 +1043,8 @@ kaa_error_t kaa_topic_list_updated(kaa_notification_manager_t *self, kaa_list_t
     }
 
     kaa_list_destroy(self->topics, &destroy_topic);
+    kaa_list_sort(new_topics,&sort_topic_by_id);
```
ohamov: Code style: please add space between arguments

client/client-multi/client-c/src/kaa/kaa_notification_manager.c
```diff
@@ -1026,6 +1043,8 @@ kaa_error_t kaa_topic_list_updated(kaa_notification_manager_t *self, kaa_list_t
...
+    self->topic_list_hash = kaa_list_hash(new_topics,&get_topic_id);
```
ohamov: Ditto

client/client-multi/client-c/src/kaa/kaa_notification_manager.c
aburdeinyi: Fixed

client/client-multi/client-c/src/kaa/collections/kaa_list.c
aburdeinyi, line 321: This is exact copy of the java hash algorithm for array of longs. This constant can be estimated from size of long and int types. I think all this constants should be used as is because it should report same output as java function.

forGGe, line 321: Using int32 here will result in UB when it comes to integer overflow. It is easy to hit due to multiplication on every iteration.
See
https://www.securecoding.cert.org/confluence/display/c/INT32-C.+Ensure+that+operations+on+signed+integers+do+not+result+in+overflow

In close-term practice this may force to fail any instrumentation (like -ftrapv flag or runtime-checking tools) that designed to catch integer overflows.

client/client-multi/client-c/src/kaa/kaa_notification_manager.c
```diff
@@ -123,6 +124,21 @@ static bool find_notifications_by_topic(void *data, void *context)
     return node->topic_id == *((uint64_t *) context);
 }
 
+static bool sort_topic_by_id(void *node_1, void *node_2)
+{
+    KAA_RETURN_IF_NIL2(node_1, node_2, false);
+    kaa_topic_t *wrapper_1 = (kaa_topic_t *)node_1;
+    kaa_topic_t *wrapper_2 = (kaa_topic_t *)node_2;
```
forGGe, line 131: Not an error, but rather recommendation - avoid explicit casting to\from void \* pointers since implicit conversion is allowed by C standard and it is type-safe as well. C99 6.3.2.3 Pointers

client/client-multi/client-c/src/kaa/kaa_notification_manager.c
```diff
@@ -123,6 +124,21 @@ static bool find_notifications_by_topic(void *data, void *context)
...
+    return wrapper_1->id < wrapper_2->id;
+}
+
+static uint64_t get_topic_id(void *node)
+{
+    //KAA_RETURN_IF_NIL(node, 0);
```
forGGe, line 137: Do not leave commented code

client/client-multi/client-c/test/collections/test_kaa_list.c
```diff
@@ -26,6 +26,26 @@
 
 static kaa_logger_t *logger = NULL;
 
+typedef struct {
+    uint64_t id;
+} test_list_node_t;
+
+static bool test_kaa_list_predicate(void *node_1, void *node_2)
+{
+    //KAA_RETURN_IF_NIL2(node_1, node_2, false);
```
forGGe: Do not leave commented code

client/client-multi/client-c/test/collections/test_kaa_list.c
```diff
@@ -26,6 +26,26 @@
...
+    test_list_node_t *wrapper_1 = (test_list_node_t *)node_1;
+    test_list_node_t *wrapper_2 = (test_list_node_t *)node_2;
+    return wrapper_1->id < wrapper_2->id;
+}
+
+static uint64_t test_kaa_list_hash(void *node)
+{
+    //KAA_RETURN_IF_NIL2(node_1, node_2, false);
```
forGGe: Do not leave commented code

client/client-multi/client-c/test/collections/test_kaa_list.c
```diff
@@ -94,6 +114,68 @@ void test_list_push_back()
...
+void test_list_hash()
+{
+    KAA_TRACE_IN(logger);
+
+    kaa_list_t *list = kaa_list_create();
+    ASSERT_NOT_NULL(list);
+
+    test_list_node_t *node;
+    uint64_t node_number = 100;
+    //TODO Initialize list with known data
```
forGGe: Should this TODO be addressed?

client/client-multi/client-c/test/collections/test_kaa_list.c
```diff
@@ -94,6 +114,68 @@ void test_list_push_back()
...
+    for (uint64_t i = 0; i < node_number; ++i) {
+        node = (test_list_node_t *)KAA_MALLOC(sizeof(test_list_node_t));
+        ASSERT_NOT_NULL(node);
+        //node->id = (uint64_t) random();
```
forGGe: Do not leave commented code.

client/client-multi/client-c/src/kaa/kaa_notification_manager.c
aburdeinyi, line 131: I think this is more related to kaa_list C implementation, and we need task one to fix this. Because kaa_list use void\* pointer to save data and don't have any information about data type. Current implementation require callbacks definition at place where list should be used for handling node's type for sorting, foreach loop and hash estimation.

client/client-multi/client-c/src/kaa/collections/kaa_list.c
aburdeinyi, line 321: Java integer overflow behaviour for this case are binary equal to C uint, because for Java it wraps around (Integer.MAX_VALUE +1 == Integer.MIN_VALUE) and for unit32_t (0x7fffffff+1 == 0x80000000). So we estimate it in uint32 but convert to int32 to be compatible with other clients.

forGGe:
Looks good to me

PR #459, opened 2016-02-01
KAA-834: Forbade to create application with same name within one tenant

vpogorelovmapr:


server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/model/sql/Application.java
```diff
@@ -34,7 +35,8 @@
 import static org.kaaproject.kaa.server.common.dao.model.sql.ModelUtils.getLongId;
 
 @Entity
-@Table(name = APPLICATION_TABLE_NAME)
+@Table(name = APPLICATION_TABLE_NAME, 
+        uniqueConstraints = { @UniqueConstraint( columnNames = {APPLICATION_TENANT_ID, APPLICATION_NAME })})
```
rasendubi: alignment

PR #468, opened 2016-02-08
KAA-839: Use CMAKE_INSTALL_PREFIX instead of KAA_INSTALL_PATH

RostakaGmfun:


rasendubi:
LGTM

client/client-multi/client-cpp/CMakeLists.txt
```diff
@@ -318,13 +318,8 @@ else (WIN32)
...
-endif()
 
-message(STATUS "KAA WILL BE INSTALLED TO  ${KAA_INSTALL_PATH} ")
-set(CMAKE_INSTALL_PREFIX ${KAA_INSTALL_PATH} ) 
-install (DIRECTORY kaa/ DESTINATION ${KAA_INSTALL_PATH}/include/kaa
+message(STATUS "KAA WILL BE INSTALLED TO ${CMAKE_INSTALL_PREFIX} ")
+install (DIRECTORY kaa/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include/kaa
     FILES_MATCHING PATTERN *.hpp)
-install (TARGETS kaacpp kaacpp_s DESTINATION ${KAA_INSTALL_PATH}/lib)
+install (TARGETS kaacpp kaacpp_s DESTINATION ${CMAKE_INSTALL_PREFIX}/lib)
```
forGGe: CMAKE_INSTALL_PREFIX do not need to be used explicitly, see:
https://cmake.org/cmake/help/v3.0/command/install.html 

> DESTINATION
>     Specify the directory on disk to which a file will be installed. If a full path (with a leading slash or drive letter) is given it is used directly. If a relative path is given it is interpreted relative to the value of the CMAKE_INSTALL_PREFIX variable. The prefix can be relocated at install time using the DESTDIR mechanism explained in the CMAKE_INSTALL_PREFIX variable documentation

rasendubi: Please, remove `${CMAKE_INSTALL_PREFIX}` from destination and use relative path. e.g. `install (TARGETS kaacpp kaacpp_s DESTINATION lib)`

forGGe:
Reviewed

PR #473, opened 2016-02-09
KAA-843: Use CMAKE_INSTALL_PREFIX intead of KAA_INSTALL_PATH

RostakaGmfun:


client/client-multi/client-c/CMakeLists.txt
```diff
@@ -174,21 +174,12 @@ if(NOT DEFINED KAA_BUILD_STATIC_ONLY)
...
-    else()
-	    set(KAA_INSTALL_PATH "/usr/local")
-    endif()
-endif()
-
-message("KAA WILL BE INSTALLED TO  ${KAA_INSTALL_PATH} ")
-set(CMAKE_INSTALL_PREFIX ${KAA_INSTALL_PATH} ) 
-install (DIRECTORY ${KAA_SRC_FOLDER}/ DESTINATION ${KAA_INSTALL_PATH}/include/kaa
+message("KAA WILL BE INSTALLED TO  ${CMAKE_INSTALL_PREFIX} ")
+set(CMAKE_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX} )
```
ghost: Do we need this?

client/client-multi/client-c/CMakeLists.txt
```diff
@@ -174,21 +174,12 @@ if(NOT DEFINED KAA_BUILD_STATIC_ONLY)
...
-message("KAA WILL BE INSTALLED TO  ${KAA_INSTALL_PATH} ")
-set(CMAKE_INSTALL_PREFIX ${KAA_INSTALL_PATH} ) 
-install (DIRECTORY ${KAA_SRC_FOLDER}/ DESTINATION ${KAA_INSTALL_PATH}/include/kaa
+message("KAA WILL BE INSTALLED TO  ${CMAKE_INSTALL_PREFIX} ")
+set(CMAKE_INSTALL_PREFIX ${CMAKE_INSTALL_PREFIX} ) 
+install (DIRECTORY ${KAA_SRC_FOLDER}/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include/kaa
     FILES_MATCHING PATTERN *.h)
 
-install (TARGETS ${KAA_BUILD_SHARED} ${KAA_BUILD_STATIC} DESTINATION ${KAA_INSTALL_PATH}/lib)
+install (TARGETS ${KAA_BUILD_SHARED} ${KAA_BUILD_STATIC} DESTINATION ${CMAKE_INSTALL_PREFIX}/lib)
```
forGGe: CMAKE_INSTALL_PREFIX do not need to be used explicitly, see:
https://cmake.org/cmake/help/v3.0/command/install.html 

> DESTINATION
>     Specify the directory on disk to which a file will be installed. If a full path (with a leading slash or drive letter) is given it is used directly. If a relative path is given it is interpreted relative to the value of the CMAKE_INSTALL_PREFIX variable. The prefix can be relocated at install time using the DESTDIR mechanism explained in the CMAKE_INSTALL_PREFIX variable documentation

forGGe:
Reviewed

PR #486, opened 2016-02-17
KAA-860: fixed persisting topic list hash

abohomol:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/persistence/KaaClientPropertiesState.java
```diff
@@ -30,12 +30,7 @@
 import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.security.PublicKey;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Properties;
-import java.util.UUID;
+import java.util.*;
```
ashvayka: Please ALWAYS use explicit imports.

ashvayka:
Minor comment, otherwise LGTM

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/persistence/KaaClientPropertiesState.java
abohomol: Got it, done.

PR #504, opened 2016-02-29
[KAA-416] Integrate Kaa with Docker

sergeyevd:
Provide two-node Docker-compose example.

### DIFF
--- a/server/containers/docker/docker-compose.yaml
+++ b/server/containers/docker/docker-compose.yaml
@@ -1,0 +1,57 @@
+cassandrahost:
+    image: spotify/cassandra
+    ports:
+        - "9042:9042"
+        - "9160:9160"
+    stdin_open: false
+zookeeperhost:
+    image: jplock/zookeeper
+    ports:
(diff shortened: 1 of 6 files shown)


server/containers/docker/docker-compose.yaml
```diff
@@ -0,0 +1,57 @@
...
+        - MONGODB_HOSTS=mongodbhost:27017
+        - JDBC_HOST=postgreshost
+        - JDBC_PORT=5432
+    command: bash -c "/config_listener.sh && /usr/bin/kaa-node start && while true; do sleep 1; done"
+    links:
+        - cassandrahost
+        - zookeeperhost
+        - mongodbhost
+        - postgreshost
+kaa_node2:
```
ashvayka: why there is two nodes?

sergeyevd: Just example of clustering. We may change it to one or three.

rasendubi:
Please, reopen against `develop`

PR #519, opened 2016-03-09
KAA-857: Use CMAKE_BUILD_TYPE instead of KAA_DEBUG_ENABLED

RostakaGmfun:


client/client-multi/client-cpp/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ if (NOT MSVC)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
 endif ()
 
-if (KAA_DEBUG_ENABLED)
+if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
```
forGGe, line 61: There is no need to check a build type, since CMake already provides set of properties that are included by CMake itself:
https://cmake.org/cmake/help/v3.0/variable/CMAKE_BUILD_TYPE.html

>  There are many per-config properties and variables (usually following clean SOME_VAR_<CONFIG> order conventions), such as CMAKE_C_FLAGS_<CONFIG>, specified as uppercase: CMAKE_C_FLAGS_[DEBUG|RELEASE|RELWITHDEBINFO|MINSIZEREL]. For example, in a build tree configured to build type Debug, CMake will see to having CMAKE_C_FLAGS_DEBUG settings get added to the CMAKE_C_FLAGS settings. See also CMAKE_CONFIGURATION_TYPES.

forGGe:
Postponed

PR #532, opened 2016-03-11
KAA-699 Cross-compilation issues addressed (SDK part)

forGGe:
Corresponding PR in sample-apps repo is here: https://github.com/kaaproject/sample-apps/pull/60

### DIFF
--- a/client/client-multi/client-c/CMakeLists.txt
+++ b/client/client-multi/client-c/CMakeLists.txt
@@ -16,14 +16,14 @@
 
 cmake_minimum_required(VERSION 2.8.8)
 
-project (Kaa-c)
+project(Kaa-c)
 enable_language(C)
 
 # Path to CMake's scripts.
-set(CMAKE_MODULE_PATH 
-        ${CMAKE_MODULE_PATH} 
+set(CMAKE_MODULE_PATH
+        ${CMAKE_MODULE_PATH}
         ${CMAKE_CURRENT_SOURCE_DIR}/Modules
-    )
+        )
 
 #
 # Sets common compilation flags.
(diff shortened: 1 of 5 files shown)


client/client-multi/client-c/toolchains/cc32xx.cmake
```diff
@@ -16,21 +16,78 @@
...
-  set(KAA_ARM_TOOLCHAIN "/opt/kaa/gcc-arm-none-eabi")
+set(CMAKE_SYSTEM_NAME Generic)
+
+set(ARM_GCC_COMPILER "arm-none-eabi-gcc${CMAKE_EXECUTABLE_SUFFIX}")
+
+# Find toolchain path
+
+if (NOT DEFINED ENV{CC32XX_TOOLCHAIN_PATH})
+  # Check if GCC is reachable.
+  find_path(TOOLCHAIN_PATH bin/${ARM_GCC_COMPILER})
```
rasendubi, line 27: You'd better search for `arm-none-eabi-gcc` directly (not `bin/...`)

client/client-multi/client-c/toolchains/cc32xx.cmake
```diff
@@ -16,21 +16,78 @@
...
+
+  if (NOT TOOLCHAIN_PATH)
```
rasendubi, line 29: I think it's safe to just error in this case.

forGGe, line 29: This will break original build behavior: if toolchain path wasn't set, it must be defaulted to /opt/kaa/gcc-arm-none-eabi.  
[source](https://github.com/kaaproject/sample-apps/blob/master/configurationdemo/source/cc32xx/platform/cc32xx/cc32xx.cmake)

It is worth to review it future released but now it will break build interfaces.

client/client-multi/client-c/toolchains/cc32xx.cmake
forGGe, line 27: Original idea was to use TOOLCHAIN_PATH pointing to dir containing whole toolchain, not only binaries:
[source](https://github.com/kaaproject/sample-apps/blob/master/configurationdemo/source/cc32xx/platform/cc32xx/cc32xx.cmake)

But I'm fine to change it, this will not break anything I hope.

client/client-multi/client-c/CMakeLists.txt
```diff
@@ -16,56 +16,59 @@
...
+        )
 
 #
 # Sets common compilation flags.
-# 
-# NOTE: Platform specific compilation flags should be set 
+#
+# NOTE: Platform specific compilation flags should be set
 # in the corresponding platform CMake script(s).
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fPIC -Wall")
```
ghost, line 36: Could you move the "-fPIC" flag to the CMAKE_COMPILER_IS_GNUCC part?
Or maybe it is better to open a new ticket.

rasendubi, line 36: Why should it be moved? The flag still should be applied even if a compiler is clang.

ghost, line 36: Well, yes. 
This flag is absent in MSVC, but Kaa 0.8.0 C SDK doesn't support this compiler at all, so, as I wrote before, we can leave this flag but open to feature story to add support of MSVC.

rasendubi:
LGTM

client/client-multi/client-c/listfiles/platform/cc32xx/CMakeLists.txt
```diff
@@ -29,31 +29,31 @@ set(KAA_SOURCE_FILES
...
-                   ${CC32XX_SDK}/simplelink/include
-                   ${CC32XX_SDK}/simplelink/source
-                   ${CC32XX_SDK}/example/common
-   )
+        ${CC32XX_SDK}/driverlib
+        ${CC32XX_SDK}/oslib
+        ${CC32XX_SDK}/simplelink
+        ${CC32XX_SDK}/simplelink/include
+        ${CC32XX_SDK}/simplelink/source
+        ${CC32XX_SDK}/example/common
```
rasendubi, line 55: Doesn't build with cc3200-sdk-1.2.0. Add `${CC32XX_SDK}/simplelink_extlib/provisioninglib`

rasendubi:
LGTM

PR #540, opened 2016-03-14
KAA-820 Stop build script when error occurs

forGGe:
Invoking of different tools are protected, so
they will not be mistakenely triggered if they are not installed.


rasendubi:
LGTM

ashvayka:
why master?

forGGe:
PR will target 0.8.2 branch when it will be done.

forGGe:
Moved to: https://github.com/rasendubi/kaa/pull/1

PR #549, opened 2016-03-21
KAA-889: Ability to use MariaDB as a SQL database in Kaa cluster

v-chyzhevskyi:


server/common/dao/pom.xml
```diff
@@ -152,7 +152,6 @@
 		<dependency>
 			<groupId>org.mariadb.jdbc</groupId>
 			<artifactId>mariadb-java-client</artifactId>
-			<scope>test</scope>
```
ashvayka, line 155: why you have removed this?

server/node/pom.xml
```diff
@@ -270,6 +270,16 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.mariadb.jdbc</groupId>
+            <artifactId>mariadb-java-client</artifactId>
+            <scope>test</scope>
```
ashvayka, line 275: why do you need this here?

PR #553, opened 2016-03-24
Add ability to create unicast notifications via Web UI

DimaMir999:
If clientKeyHash field is empty - send broadcast notification else  unicast notification

### DIFF
--- a/server/node/src/main/resources/org/kaaproject/kaa/server/admin/client/i18n/KaaAdminConstants.properties
+++ b/server/node/src/main/resources/org/kaaproject/kaa/server/admin/client/i18n/KaaAdminConstants.properties
@@ -109,6 +109,8 @@ className = Class name
 
 classType = Class type
 
+clientKeyHash = Client key hash (for unicast notification)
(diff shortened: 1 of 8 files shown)


server/node/src/main/resources/org/kaaproject/kaa/server/admin/client/i18n/KaaAdminConstants.properties
```diff
@@ -109,6 +109,8 @@ className = Class name
 
 classType = Class type
 
+clientKeyHash = Client key hash (for unicast notification)
```
ashvayka: please change the label to "Endpoint ID"

ashvayka: please change clientKeyHash to endpointKeyHash in the code.

DimaMir999:
https://github.com/kaaproject/kaa/pull/556
my pull request in correct branch

PR #554, opened 2016-03-24
KAA-924 True Metal sync

rasendubi:


forGGe:
Looks good (already reviewed, actually)

PR #562, opened 2016-03-30
Fixed calculation of new_buffer_size

eudyptula:
While sending events over Kaa, we encountered problems with larger message sizes. I tracked the issue to the calculation of new_buffer_size that seems to be off.

The attached solution solves the problem and I thought you might be interested in including this.

Only tested briefly on POSIX where realloc takes the new total size as parameter - not sure if it works on all the other of your supported platforms.


rasendubi:
Hi, @eudyptula. Could you please add a test case showing the issue to the test suite?

The change looks good to me, but I'll take a second look at `kaa_buffer.c`. The file seems a bit messy to me—it may need rewriting.

Oh, and the Travis have failed. (`free_space` variable is now unused and should be removed.)

eudyptula:
I'm not so familiar with your test setup, but I added a few lines that would have caught the original problem.

rasendubi:
Many thanks! LGTM

cc @forGGe

forGGe:
Looks good to me

rasendubi:
Thanks for your contribution!

PR #565, opened 2016-03-30
KAA-896: added ability to set failover strategy

abohomol:


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/failover/strategies/FailoverStrategy.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.kaaproject.kaa.client.channel.failover.FailoverDecision;
+import org.kaaproject.kaa.client.channel.failover.FailoverStatus;
+
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Failover strategy is responsible for producing failover decisions based on failover statuses.
+ */
+public interface FailoverStrategy {
```
ashvayka, line 28: There is no method that notifies FailoverStrategy that SDK recovered from the failure. This makes it impossible to implement more or less production ready scenario: "If there is a failure to connect to bootstrap servers i will retry it for 3 times and then notify user somehow". Basically you know that there was subsequent failures but you don't know if you ever recovered between failures.

client/client-multi/client-java-core/src/test/java/org/kaaproject/kaa/client/channel/failover/DefaultFailoverManagerTest.java
```diff
@@ -44,13 +46,15 @@
...
     private ExecutorContext context;
+    private FailoverStrategy failoverStrategy;
 
     @Before
     public void setUp() {
         channelManager = Mockito.mock(KaaChannelManager.class);
         context = Mockito.mock(ExecutorContext.class);
         Mockito.when(context.getScheduledExecutor()).thenReturn(Executors.newScheduledThreadPool(1));
-        failoverManager = new DefaultFailoverManager(channelManager, context, RESOLUTION_TIMEOUT_MS, BOOTSTRAP_RETRY_PERIOD, 1, 1, TimeUnit.MILLISECONDS);
+        failoverStrategy = Mockito.spy(new DefaultFailoverStrategy(BOOTSTRAP_RETRY_PERIOD, 1, 1, TimeUnit.MILLISECONDS));
```
ashvayka: How one can define failover strategy in real time, for example before client initialization?

ashvayka:
There are some major usability issues that need to be resolved before this PR can be merged

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/failover/strategies/FailoverStrategy.java
abohomol, line 28: FailoverManager is aware of recovery so it's possible to let FailoverManager notify strategy about that. We could also pass connection information to strategy.
Let me know if it's what you've been looking for.

client/client-multi/client-java-core/src/test/java/org/kaaproject/kaa/client/channel/failover/DefaultFailoverManagerTest.java
abohomol: Added ability to switch failover strategy at runtime: https://github.com/kaaproject/kaa/pull/565/commits/a7bbbc993861ec20e568968ad56a6be21c3e0dcd

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/failover/strategies/FailoverStrategy.java
abohomol, line 28: [Updated](https://github.com/kaaproject/kaa/pull/565/commits/e77079a407ca6ab1fd0f5a3fd45dbf79903d9545).

ashvayka, line 28: LGTM

client/client-multi/client-java-core/src/test/java/org/kaaproject/kaa/client/channel/failover/DefaultFailoverManagerTest.java
ashvayka: Require dedicated API call to set strategy from KaaClient interface

ashvayka:
@dyosick please review

ashvayka:
LGTM :+1:

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/failover/strategies/DefaultFailoverStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+        this.noConnectivityRetryPeriod = noConnectivityRetryPeriod;
+        this.timeUnit = timeUnit;
+    }
+
+    @Override
+    public FailoverDecision onFailover(FailoverStatus failoverStatus) {
+        LOG.trace("Producing failover decision for failover status: {}", failoverStatus);
+        switch (failoverStatus) {
+            case BOOTSTRAP_SERVERS_NA:
+                return new FailoverDecision(FailoverAction.RETRY, bootstrapServersRetryPeriod, timeUnit);
```
ghost, line 67: I think the better way is to switch to the first one.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/failover/strategies/DefaultFailoverStrategy.java
```diff
@@ -0,0 +1,100 @@
...
+            case CURRENT_BOOTSTRAP_SERVER_NA:
+                return new FailoverDecision(FailoverAction.USE_NEXT_BOOTSTRAP, bootstrapServersRetryPeriod, timeUnit);
+            case NO_OPERATION_SERVERS_RECEIVED:
+                return new FailoverDecision(FailoverAction.USE_NEXT_BOOTSTRAP, bootstrapServersRetryPeriod, timeUnit);
```
dyosick: As I understood, we postpone switching to the next bootstrap server. I thing we can send a request immediately.

ghost, line 71: As I understood, we postpone switching to the next bootstrap server. I think we can send a request immediately.

client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/channel/failover/strategies/DefaultFailoverStrategy.java
abohomol, line 67: Both Java and C++ clients produce same decision with same retry period. Are we going to change it everywhere? Same for comment below.
Also there is no action to select first bootstrap - I assume it must be responsibility of Channel manager or exact channel to choose.

ghost, line 67: I think we need to create one item (for all SDKs) for the OPERATION_SERVERS_NA status. I suggest to use the USE_NEXT_BOOTSTRAP action instead of RETRY in this case.

And another item (for all SDKs) is for the BOOTSTRAP_SERVERS_NA action to use USE_NEXT_BOOTSTRAP instead, because RETRY means action with a current server, but this action is not acceptable, I think, in case when this server doesn't respond.

abohomol, line 67: I agree, but we may also consider creating separate story for changing default failover behaviour as well as implementing few custom ones like with LogUploadStrategy. Does it make sense?

rasendubi:
@dyosick left LGTM at the ticket.

PR #569, opened 2016-03-31
KAA-950: fix Kaa client state persistence

abohomol:


sergiyk0:
Looks good to me

am1kk:
LGTM.

PR #570, opened 2016-03-31
Kaa 926 Make KAA_MAX_LOG_LEVEL overridable for tests

Dubland:
/cc @rasendubi @forGGe

### DIFF
--- a/client/client-multi/client-c/build.sh
+++ b/client/client-multi/client-c/build.sh
@@ -30,9 +30,9 @@ then
     help
 fi
 
-if [ -z ${MAX_LOG_LEVEL} ]
+if [ -z ${MAX_LOG_LEVEL+x} ]
 then
-    MAX_LOG_LEVEL=6;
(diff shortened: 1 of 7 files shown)


client/client-multi/client-c/build.sh
```diff
@@ -30,9 +30,13 @@ then
     help
 fi
 
+if [ -z ${MAX_LOG_LEVEL} ]
```
rasendubi: The following solution is incorrect: http://stackoverflow.com/a/13864829/2538771

client/client-multi/client-c/build.sh
```diff
@@ -30,9 +30,13 @@ then
...
+then
+    MAX_LOG_LEVEL=6;
```
rasendubi: `;` is not necessary

rasendubi:
LGTM

forGGe:
Looks good

PR #575, opened 2016-04-04
KAA-942: Deep schema export fails with CTL dependencies

bkhablenko:
The dependencies were searched for using the parent schema's application ID only, thus making it impossible to reference any common types within the tenant scope.

### DIFF
--- a/server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/service/CTLServiceImpl.java
+++ b/server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/service/CTLServiceImpl.java
@@ -553,13 +553,8 @@ public class CTLServiceImpl implements CTLService {
                 ObjectNode dependency = (ObjectNode) node;
                 String fqn = dependency.get(FQN).getTextValue();
                 Integer version = dependency.get(VERSION).getIntValue();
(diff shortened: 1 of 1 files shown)


server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/service/CTLServiceImpl.java
```diff
@@ -554,6 +555,12 @@ private Schema parseDependencies(CTLSchemaDto schema, final Schema.Parser parser
                 Integer version = dependency.get(VERSION).getIntValue();
                 CTLSchemaDto child = this.findCTLSchemaByFqnAndVerAndTenantIdAndApplicationId(
                         fqn, version, parent.getMetaInfo().getTenantId(), parent.getMetaInfo().getApplicationId());
+                if (child == null) { // Not found within the application scope
+                    String tenantId = parent.getMetaInfo().getTenantId();
+                    String applicationId = null; // Search within the tenant scope then
+                    child = this.findAnyCTLSchemaByFqnAndVerAndTenantIdAndApplicationId(fqn, version, tenantId, applicationId);
```
ashvayka: I believe that you can use thins method instead of "findCTLSchemaByFqnAndVerAndTenantIdAndApplicationId" and there will be no need to execute request twise (worst case scenario).

ashvayka:
Since we have reviewed this with @ikulikov, LGTM

PR #591, opened 2016-04-08
KAA-993 Add dec2hex converter to handle window's versions 10 and later

aburdeinyi:
added dec2hex converter to handle window's versions 10


ghost:
LGTM

rasendubi:
LGTM

PR #593, opened 2016-04-11
KAA-976: Review and update REST API calls to accept application token instead of application ids.

sashadidukh:
Excluding CTL REST API calls.


rasendubi:
add to whitelist

server/common/admin-rest-client/src/main/java/org/kaaproject/kaa/server/common/admin/AdminClient.java
```diff
@@ -305,6 +305,14 @@ public TopicDto getTopic(String topicId) throws Exception {
         return entity.getBody();
     }
 
+    public List<TopicDto> getTopicsByApplicationToken(String applicationToken) throws Exception {
```
ashvayka, line 312: All corresponding methods that use applicationId should be marked as Deprecated and contain reference to corresponding new method

PR #602, opened 2016-04-12
KAA-947 Warning fix

forGGe:


forGGe:
cc @dyosick @rasendubi

rasendubi:
LGTM

PR #615, opened 2016-04-15
KAA-1015: Update default record generation algorithm to process integral numeric values for integer and long avro field types

ikulikov:


iuriigolovach:

PR #616, opened 2016-04-15
KAA-876: Method for revoking of credentials fixed

v-chyzhevskyi:
Method for revoking of credentials fixed


iuriigolovach:

PR #620, opened 2016-04-18
KAA-1017 Make build-posix default directory

rasendubi:

PR #621, opened 2016-04-18
KAA-998 Add build without logs

rasendubi:
Done on top of #620 to avoid merge conflicts.

/cc @forGGe


iuriigolovach:

forGGe:
+

PR #628, opened 2016-04-19
KAA-1021 Update Nix packages

rasendubi:
- gcc-arm-embedded-5_2 is already in channel
- merged avro-cpp to upstream, waiting for channel update
- submitted astyle patch to upstream


forGGe:
+

PR #634, opened 2016-04-21
KAA-1024: Set default server log level to WARN

krupet:


kaa-leeroy:

rasendubi:
@kaa-leeroy add to whitelist

ashvayka:
Usually, server log level should be set to INFO, why we are setting to WARN?

akokhanovskyi:
@krupet @ashvayka Can we drop this PR? I think we agreed that the appropriate way to solve the issue at hand is by reviewing logging levels in general.

rasendubi:
Closing this. If still needed, feel free to reopen.

PR #640, opened 2016-04-21
KAA-965 Change default behaviour for new device management codes

Dubland:
/cc @rasendubi @forGGe


forGGe:
+

rasendubi:

PR #650, opened 2016-04-22
KAA-874: Added getConfigurationRecordBody API call

arturjoshi:


kaa-leeroy:

rasendubi:
@kaa-leeroy add to whitelist

ikulikov:
@arturjoshi - getConfigurationRecordBody API call should be also added to [AdminClient](https://github.com/kaaproject/kaa/blob/develop/server/common/admin-rest-client/src/main/java/org/kaaproject/kaa/server/common/admin/AdminClient.java).

PR #651, opened 2016-04-23
[KAA-1027] Endpoint SDK should not control application behaviour

abohomol:
Also this pull request contains commit with command to suppress warnings in Pods.


rasendubi:

PR #652, opened 2016-04-25
KAA-951: Change cassandra client connection port number in order to be able to connect to test cassandra cluster.

ikulikov:


kaa-leeroy:

rasendubi:
@kaa-leeroy add to whitelist

PR #660, opened 2016-04-27
KAA-915: Expose the existing metrics to JMX

bkhablenko:
http://jira.kaaproject.org/browse/KAA-915


kaa-leeroy:

rasendubi:
@kaa-leeroy add to whitelist

PR #666, opened 2016-05-08
KAA-1057 Switch to upstream avro-cpp

rasendubi:
avro-cpp nix-expression is now available in the unstable nixpkg channel,
so it's better to use it.

/cc @forGGe


forGGe:
looks good

PR #669, opened 2016-05-16
KAA-1080 Use travis_retry to reduce impact of network failure

rasendubi:


rasendubi:

PR #677, opened 2016-05-17
KAA-1070 Update doxygen configuration

rasendubi:
Doxygen complained about obsolete options, so I just ran `doxygen -U`.
I've also cleaned up doxygen cmake module a little.

/cc @GamovCoder


forGGe:
Looks good

PR #678, opened 2016-05-18
Kaa 1090 Fix bug with client reconnection after network drop

am1kk:


kaa-leeroy:

PR #680, opened 2016-05-18
Bug/kaa 1092

Kirill380:


kaa-leeroy:

PR #693, opened 2016-05-24
KAA-1031: Group/Filter lookup Failure during process of Endpoint request

Kirill380:


kaa-leeroy:

rasendubi:
@kaa-leeroy add to whitelist

server/node/src/main/java/org/kaaproject/kaa/server/operations/service/cache/concurrent/ConcurrentCacheService.java
```diff
@@ -323,7 +323,7 @@ public String putConfId(ConfigurationIdKey key, String value) {
                         relatedChanges.add(historyDto);
                     } else if (changeType == ChangeType.ADD_PROF || changeType == ChangeType.REMOVE_PROF) {
                         ProfileFilterDto profileFilter = profileService.findProfileFilterById(changeDto.getProfileFilterId());
-                        if (supports(profileFilter, key.getEndpointProfileSchemaVersion(), key.getServerProfileSchemaVersion())) {
+                        if (profileFilter != null && supports(profileFilter, key.getEndpointProfileSchemaVersion(), key.getServerProfileSchemaVersion())) {
```
ashvayka, line 326: This is not valid. We can't simply ignore the fact that profileFilter was deleted. Based on the result of this method, we calculate affect on endpoints. 
This particular filter may or may not be related to our endpoint. "supports" method is responsible for figuring this out.
However, support method requires client and server side EP profile schema versions.

I believe that we need to change way how we treat deleted profile filter. We need to mark them as deleted and 
In case changeType == ChangeType.ADD_PROF we can ignore deleted profile filter.
In case changeType == ChangeType.REMOVE_PROF we need to fetch deleted filter using special method and invoke "supports" method.

ashvayka, line 326: You can either introduce "deleted" flag for profile filters or copy this data to other table. I would use "deleted" flag.

ashvayka:
PR name should contain ticket description: 
KAA-1031: Group/Filter lookup Failure during process of Endpoint request

PR #695, opened 2016-05-26
KAA-1051: PostgreSQL database driver not found exception while Kaa 0.9.0 deb installation

v-chyzhevskyi:
Added database drivers of PostgreSQL and MariaDB to general dependencies in kaa-node module


rasendubi:
You should branch off from the `develop` branch.

PR #697, opened 2016-05-26
KAA-1030: Ability to delete endpoints

v-chyzhevskyi:
Added ability to delete endpoints (both through GUI and REST).


rasendubi:
Branch off from the `develop` branch.

PR #700, opened 2016-05-26
KAA-997 Add Nix guide

rasendubi:
/cc @forGGe

### DIFF
--- a/client/client-multi/client-c/docs/nix.org
+++ b/client/client-multi/client-c/docs/nix.org
@@ -1,0 +1,60 @@
+#+TITLE: Nix Guide
+#+AUTHOR: Alexey Shmalko
+#+OPTIONS: toc:nil
+
+[[http://nixos.org/nix/][Nix]] is a powerful Linux package manager. We use it to create a better development environment for C client and manage all dependencies.
+
+* Setup Nix
+  The easiest way to setup Nix is next:
+  #+begin_src sh
+    curl https://nixos.org/nix/install | sh
+    source $HOME/.nix-profile/etc/profile.d/nix.sh
+  #+end_src
+
+  And add =source $HOME/.nix-profile/et/profile.d/nix.sh= to your =.bashrc=.
+
+  If you don't trust piping shell scripts from the Internet (and you shouldn't), feel free to examine the script or use alternate setup. (Though, you should google it up yourself.)
+
+* Install all dependencies and enter the shell
+  The first time you enter shell environment, the Nix will install all dependencies needed for development.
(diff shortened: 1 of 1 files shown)


client/client-multi/client-c/docs/nix.org
```diff
@@ -0,0 +1,60 @@
...
+  If you want to run a single command within a shell, you can use =--run= option. For example:
+  #+begin_src sh
+    nix-shell --run make
+  #+end_src
+
+* Options
+  | Option             | Default value | Meaning                                        |
+  |--------------------+---------------+------------------------------------------------|
+  | posixSupport       | true          | Host build with gcc. Goes to =build-posix=.    |
+  | clangSupport       | true          | Host build with clang. Goest to =build-clang=. |
```
rasendubi: Typo: `Goest`

client/client-multi/client-c/docs/nix.org
```diff
@@ -0,0 +1,60 @@
...
+  | withWerror         | false         | Enable =-Werror= for all builds.               |
+
+  You can override any option with the following command:
+  #+begin_src sh
+    nix-shell --arg optionName value
+  #+end_src
+
+  For example:
+  #+begin_src sh
+    nix-shell --arg withWerror false
```
rasendubi: It's better to use `true` here, as it's not default.

client/client-multi/client-c/docs/nix.org
```diff
@@ -0,0 +1,60 @@
...
+
+  If you don't trust piping shell scripts from the Internet (and you shouldn't), feel free to examine the script or use alternate setup. (Though, you should google it up yourself.)
+
+* Install all dependencies and enter the shell
+  The first time you enter shell environment, the Nix will install all dependencies needed for development.
+  #+begin_src sh
+    nix-shell
+  #+end_src
+
+  As CC3200 SDK is not freely available, nix-shell will abort and ask you to download the file manually and add it to nix-store -- follow instructions. Then re-run the command above.
```
forGGe, line 24: Probably it is worth to add here example of error and explicitly give a command that will add cc32xx SDK to the nix.

forGGe:
Reviewed.

client/client-multi/client-c/docs/nix.org
rasendubi, line 24: I don't think that's needed. Nix prints both the reason and how to fix it.

forGGe, line 24: Per discussion: agreed.

forGGe:
+

PR #701, opened 2016-05-26
KAA-997 Add sandbox updating guide

rasendubi:
/cc @forGGe


rasendubi:
The patch changes no code, so jenkins failure is not relevant.

forGGe:
+

PR #708, opened 2016-05-27
KAA-1093: Credentials service name of applications from SandBox is NULL

v-chyzhevskyi:
Default credentials service name will be "Trustful"


ashvayka:
LGTM

PR #711, opened 2016-05-31
KAA-1097:Server sends ping responses in an infinite loop if the endpoint doesn't receive it

v-chyzhevskyi:
Сache didn't work


ashvayka:
LGTM!

ashvayka:
:+1:

PR #713, opened 2016-05-31
KAA-1106:Not mandatory fields are required during changing account profile settings

v-chyzhevskyi:
Not mandatory fields are required during changing account profile settings


server/node/src/main/java/org/kaaproject/kaa/server/admin/client/mvp/view/user/UserProfileViewImpl.java
```diff
@@ -106,7 +110,10 @@ protected void resetImpl() {
 
     @Override
     protected boolean validate() {
-        return email.getValue().length()>0;
+        boolean result = firstName.getValue().length()>0;
+        result &= lastName.getValue().length()>0;
+        result &= Utils.validateEmail(email.getValue());
+        return result;
```
ghost, line 116: How about?
return !firstName.getValue().isEmpty() && !lastName.getValue().isEmpty() && Utils.validateEmail(email.getValue());

ikulikov:
LGTM

PR #719, opened 2016-06-03
List of attached endpoints contain duplicates

v-chyzhevskyi:
List of attached endpoints contain duplicates

### DIFF
--- a/server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/service/EndpointServiceImpl.java
+++ b/server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/service/EndpointServiceImpl.java
@@ -326,6 +326,9 @@ public class EndpointServiceImpl implements EndpointService {
             endpointIds = new ArrayList<>();
             endpointUser.setEndpointIds(endpointIds);
         }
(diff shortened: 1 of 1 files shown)


server/common/dao/src/main/java/org/kaaproject/kaa/server/common/dao/service/EndpointServiceImpl.java
```diff
@@ -326,6 +326,9 @@ public EndpointProfileDto attachEndpointToUser(String userExternalId, String ten
             endpointIds = new ArrayList<>();
             endpointUser.setEndpointIds(endpointIds);
         }
+        else if (endpointIds != null && endpointIds.contains(profile.getId())) {
+            LOG.warn("Endpoint is already assigned to current user {}. Unassign it first!.", profile.getEndpointUserId());
```
ashvayka: LOG.warn("Endpoint is already assigned to current user {}!", profile.getEndpointUserId());

ashvayka: Maybe we need to stop execution of this method and simply return success?

ikulikov:
LGTM

PR #723, opened 2016-06-03
KAA-1101: Cannot save/activate configuration after session timeout

v-chyzhevskyi:
Cannot save/activate configuration after session timeout


ikulikov:
LGTM

PR #728, opened 2016-06-08
KAA-991: Aggregate maven build profiles using environment variables

v-chyzhevskyi:
KAA-991: add environment parameter "appenders"


ashvayka:
This PR is incomplete, We need to add "verifiers" profile to enable all verifiers.

PR #747, opened 2016-06-13
KAA-1162 Update copyright

balldir:


kaa-leeroy:

PR #758, opened 2016-06-14
KAA-1099 Accept a JSON configuration file without the __uuid field

nocs00:


kaa-leeroy:

PR #761, opened 2016-06-15
KAA-1157 Enable syntax highlighting in kramdown config

balldir:


kaa-leeroy:

PR #766, opened 2016-06-16
KAA-1186 Add support for ruby2.0 in create_global_toc.rb

balldir:


kaa-leeroy:

rasendubi:
Works for me

PR #786, opened 2016-06-21
KAA-1169 Fix cc32xx cmake files and documentation

rasendubi:
This patch started as a review fix for https://github.com/kaaproject/kaa/pull/784 but eventually grow into larger fix for both documentation and code.

cc @forGGe


forGGe:
+

SaiPrakash14:
i followed what ever steps you given but i got error like this
please tell me solution

CMake Error at CMakeLists.txt:74 (project):
  The CMAKE_C_COMPILER:

    c:/cygwin/opt/kaa/arm-none-eabi/bin/bin/arm-none-eabi-gcc

  is not a full path to an existing compiler tool.

  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.


-- Configuring incomplete, errors occurred!
See also "C:/cygwin/home/Quantum/build/CMakeFiles/CMakeOutput.log".

PR #792, opened 2016-06-22
KAA-1155: Bug with server-side profile schema version fixed

v-chyzhevskyi:
Bug with updating of server-side profile schema version fixed


ikulikov:
LGTM

PR #812, opened 2016-06-29
added validation api to server/common module, fixed bug  with ticket …

Pyshankov:
…number : KAA-1151


kaa-leeroy:

PR #834, opened 2016-07-04
KAA-1212 Drop UDOO page for C++ SDK

forGGe:
Due to lack of differences between Linux on UDOO and regular Linux distro

/cc @rasendubi @Slash32


rasendubi:
LGTM

PR #835, opened 2016-07-04
APP-73 Set cmake-minimum_required to higher version 

Dubland:
/cc @rasendubi @forGGe


forGGe:
Travis still fails. Investigate please

forGGe:
+

PR #843, opened 2016-07-06
Kaa 1239 Support for github pages hosting.

balldir:

PR #844, opened 2016-07-06
Kaa 1239 Search for documentation

balldir:


akokhanovskyi:
@balldir Who is reviewing this?

balldir:
@akokhanovskyi I don't know a proper reviewer for this one, so I've asked in development chat and nobody volunteered as for now.
Reviewer should keep in mind that it is second part of a patch (first one https://github.com/kaaproject/kaa/pull/843)
Test version with this changes is available at https://balldir.github.io/kaa

gh-pages-stub/_includes/nav.html
```diff
@@ -17,11 +17,11 @@
...
         {% endif %}
           {% if val.url == navurl %}
-            <li class="menu-selected menu-opened"><a href="{{ val.url }}">{{ postfix }}{{ val.text }}</a></li>
+            <li class="menu-selected menu-opened"><a href="{{ site.baseurl }}{{ val.url }}">{{ postfix }}{{ val.text }}</a></li>
 		  {% elsif navurl contains val.url %}
-			<li class="menu-opened"><a href="{{ val.url }}">{{ postfix }}{{ val.text }}</a></li>
+			<li class="menu-opened"><a href="{{ site.baseurl }}{{ val.url }}">{{ postfix }}{{ val.text }}</a></li>
           {% else %}
-            <li><a href="{{ val.url }}">{{ postfix }}{{ val.text }}</a></li>
+            <li><a href="{{ site.baseurl }}{{ val.url }}">{{ postfix }}{{ val.text }}</a></li>
```
Faradey27, line 24: can this template do something like {{global.variable1 + global.variable2 }} ? or only {{global.variable1}}{{global.variable2}} ?

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
+(function() {
+
+  function isBlank(str) {
```
Faradey27: can we have something like utils and move this logic to the utils? 
i think this function can be reused in several different places

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+    return (!str || /^\s*$/.test(str));
+  } 
+
+  function displaySearchResults(results, store, searchTerm) {
+    var searchResults = document.getElementById('search-results');
```
Faradey27: maybe better to have constanst for selectors?

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+
+    if (results.length) {
+      var appendString = '';
```
Faradey27: better to prevent modification by link of the same variable

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+
+      for (var i = 0; i < results.length; i++) {
```
Faradey27: why we use for? can we use forEach or reduce here?

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+        var item = store[results[i].ref];
```
Faradey27: can we don't use global variables(like store) in displaySearchResults, maybe better to pass it as parameter

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+        if (isBlank(item.title)) {
+            item.title = "No title";
```
Faradey27: is it good idea to modify store item by link here, method name displaySearchResults, saw i beleave that this method should not modify anything

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+        }
+        appendString += '<li><a href="' + item.url + '"><h3>' + item.title + '</h3></a>';
```
Faradey27: i think better to use reduce for this

> appendString = reduce((prev, cur) => prev + '<li><a href="' + item.url + '"><h3>' + item.title + '</h3></a>')
- something like this, but a little more complex

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+	      var indx = item.content.indexOf(searchTerm);
+	      var start_pos = Math.max(0, indx - 75);
```
Faradey27: what is 75? better to not use magic numbers

Faradey27: better to write code in one style, if you use camelCase - always use it, if you use lisp_case - always use it

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+	      var stop_pos = Math.min(item.content.length, indx + 75);
+	      var str_to_add = item.content.substring(start_pos, stop_pos);
+	      str_to_add = str_to_add.replace(searchTerm, "<b><u>" + searchTerm + "</u></b>" );
```
Faradey27: modification by link - not best approach

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+        appendString += '<p>';
```
Faradey27: modification by link - not best approach

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+        if (start_pos > 0 ) {
+            appendString += '... ';
```
Faradey27: modification by link inside if - bad approach, it make much more hard to predict what will happen

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+	      }
+        appendString += str_to_add;
```
Faradey27: modification by link - not best approach

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+        if (stop_pos < item.content.length) {
```
Faradey27: same

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+           appendString += ' ...'  
+        }
+	      appendString += '</p></li>';
+      }
+      searchResults.innerHTML = appendString;
```
Faradey27: better to move logic that manipulate dom in separate method(maybe class which will control work with dom)

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+    } else {
+      searchResults.innerHTML = '<li>No results found</li>';
+    }
+  }
+
+  function getQueryVariable(variable) {
+    var query = window.location.search.substring(1);
```
Faradey27: the same with location, better to hide it and don't use directly

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+    var vars = query.split('&');
+
+    for (var i = 0; i < vars.length; i++) {
```
Faradey27: you can use map and filter here, instead of for, "for" better to use in places where we have more then 10000 items

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+        return decodeURIComponent(pair[1].replace(/\+/g, '%20'));
+      }
+    }
+  }
+
+  var searchTerm = getQueryVariable('query');
+  var searchVersion = getQueryVariable('version');
+  var searchBaseUrl = getQueryVariable('baseUrl');
+  if (searchVersion) {
+    document.getElementById("version-select").value = searchVersion;
```
Faradey27: not obvious what we do here, better to have method for this with clear name

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+  if (!searchVersion || searchVersion == "") {
```
Faradey27: we should always use === instead of ==, and also !searchVersion and searchVersion == "" - the same, they will be always true and false in same time

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+    var e = document.getElementById("version-select");
```
Faradey27: why name is e? e is common name for event

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+    searchVersion = e.options[e.selectedIndex].value;
```
Faradey27: from this line i think options is array and selecteIndex is number, [{value: it's version}, {}], can we do this more obvious?
like e.options.versionOption.value

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+  }
+  if (searchTerm) {
+    
+    ga('send', 'event', 'User Search', 'Search', searchTerm);
```
Faradey27: better to hide logic that work with analitic into some class, 
class Analytic { sendEvent(data) {ga(data)} } - something like this,
in future this will help to use other analytics or several in same time

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+    
+    document.getElementById('search-box').setAttribute("value", searchTerm);
+    
+    var url = searchBaseUrl + "/" + searchVersion + "/" + "search_dictionary.json";
```
Faradey27: better to have class in which we will create all urls

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+    $.get( url, function( data ) {
```
Faradey27: better to have class for work with api

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+          'title': data[key]["title"],
+          'content': data[key]["content"]
+        });
+      }
+      var results = idx.search(searchTerm);
+      displaySearchResults(results, data, searchTerm);
+
+    })
+    .fail(function(jqXHR, textStatus, errorThrown) {
+    alert( textStatus );
```
Faradey27: better to don't use alert, just show some modal or small notification

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+      }
+      searchResults.innerHTML = appendString;
+    } else {
+      searchResults.innerHTML = '<li>No results found</li>';
+    }
+  }
+
+  function getQueryVariable(variable) {
+    var query = window.location.search.substring(1);
+    var vars = query.split('&');
```
Faradey27: split can cause js error, better to add check that we have not empty string

gh-pages-stub/public/js/search.js
```diff
@@ -0,0 +1,93 @@
...
+
+    for (var i = 0; i < vars.length; i++) {
+      var pair = vars[i].split('=');
```
Faradey27: the same, unsafe

gh-pages-stub/_includes/nav.html
balldir, line 24: Unfortunately it doesn't work that way.

Faradey27:
LGTM

PR #863, opened 2016-07-12
KAA-1240 public host/port configuration

nocs00:


doc/Administration-guide/System-Configuration/General-configuration/index.md
```diff
@@ -4,4 +4,87 @@ title: General configuration
...
+{% include variables.md %}
+
+* TOC:
+{:toc}
+
+# Server configuration
+
+## Public host/port configuration
+
+Kaa allows you to choose what host/port to use for internal and external communication with bootstrap and operation servers.
```
ghost: The server is called Operations but not Operation.

nocs00:
Fixed

doc/Administration-guide/System-Configuration/General-configuration/index.md
```diff
@@ -4,4 +4,87 @@ title: General configuration
...
+{% include variables.md %}
+
+* TOC:
+{:toc}
+
+# Server configuration
+
+## Public host/port configuration
+
+Kaa allows you to choose what host/port to use for internal and external communication with bootstrap and operations servers.
```
ashvayka: We need to add introduction here in order to explain what is "internal" and "external" communication. Actually this is incorrect. There is bind interface and address where netty binds and serves requests and public interface which is reported to the client in form of embedded configuration in generated SDK. Need to explain why this is useful (public and private IP in cloud provider (AWS, Google Cloud, etc))

doc/Administration-guide/System-Configuration/General-configuration/index.md
```diff
@@ -4,4 +4,87 @@ title: General configuration
...
+"publicInterface":"${transport_public_interface}",
+"publicPort":9888
+}
+```
+
+### Typical usecases
+
+Given a client, which should communicate with Kaa server. Client has restrictions, for example 80 and 8000 ports are open only.
+In this case client unable to commuticate with Kaa server, because default configuration pointed to listen on different ports.
+The solution is, to change config file operations-tcp-transport.config , set *"publicPort":80* for communication with operations server via 80 port and in file bootstrap-tcp-transport.config , set *"publicPort":8000* for communication with bootstrap server via 8000 port.
```
ashvayka, line 219: This is simply not true. Please review this use case. You need to change both bind and public port. So this is not the perfect use case.

doc/Administration-guide/System-Configuration/General-configuration/index.md
```diff
@@ -19,168 +28,190 @@ $ sudo service kaa-node restart
...
 
-Interface that will be reported by all transports.
+# Interface that will be reported by all transports
+transport_public_interface=localhost
+```
 
-#### metrics_enabled
-Default: _true_
+Where *transport_bind_interface* reflects private IP for Bootstrap and Operations servers, and *transport_public_interface* - public accordingly.
+Below are given default config files, which consume above host values from file kaa-node.properties.
```
ashvayka, line 167: please add a note that if public_interface is set to localhost - SDK will send request to localhost, so only applications that are launched on the same host will work.

ashvayka:
Why this PR contains update to all configuration parameters? It is very hard to review.

ashvayka:
LGTM

PR #877, opened 2016-07-15
KAA-1266: [C++ SDK] Endpoint sometimes crashes on stop.

ghost:


client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -114,32 +119,26 @@ void ThreadPool::shutdownNow()
 void ThreadPool::start()
 {
     for (std::size_t i = 0; i < workerCount_; ++i) {
-        workers_.push_back(std::thread(Worker(*this)));
+        workers_.emplace_back(Worker(*this));
```
RostakaGmfun: The `emplace_back()` is used to push multiple values.
You can use `push_back()` here, since in C++11 this method takes rvalue references as well:

```
void push_back( T&& value );
```

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,89 +16,94 @@
...
+    for (auto &worker : workers_) {
+        if (worker.joinable()) {
+            worker.join();
+        }
+    }
 }
 
 void ThreadPool::add(const ThreadPoolTask& task)
 {
     if (!task) {
```
RostakaGmfun, line 39: Are you sure this check is needed for const reference?

RostakaGmfun:
Reviewed

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,89 +16,94 @@
...
-            task();
-        } catch (...) {}
+                task = std::move(threadPool_.tasks_.front());
+                threadPool_.tasks_.pop_front();
+            }
 
-        KAA_LOCK(tasksLock);
+            try {
+                task();
+            } catch (...) {}
```
balldir: I would prefer to log exception here.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
forGGe, line 39: This will check not a reference, but `task` itself. `ThreadPoolTask` has an overloaded `operator! ()`

RostakaGmfun, line 39: My bad, got it

forGGe:
+

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost: According to the official guide emplace_back() "constructs an element in-place at the end" (http://en.cppreference.com/w/cpp/container/list)

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost: Agree. But by design this class is not expected to have any logger. It is just like an utility class that may not have any relationships with Kaa C++ SDK.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ohamov: Denis is right, it will reduce unnecessary constructor call and destructor call, and, also, call to the move constructor.

client/client-multi/client-cpp/impl/context/AbstractExecutorContext.cpp
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include "kaa/logging/Log.hpp"
```
rasendubi, line 17: Is that needed here?

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,89 +16,94 @@
...
-
-void Worker::operator ()()
-{
-    KAA_MUTEX_UNIQUE_DECLARE(tasksLock, threadPool_.threadPoolGuard_);
-
-    while (threadPool_.isRun_) {
-        while (threadPool_.isRun_ && threadPool_.tasks_.empty() && !threadPool_.isPendingShutdown_) {
-            KAA_CONDITION_WAIT(threadPool_.onNewTask_, tasksLock);
-        }
+    void operator() ()
```
rasendubi: Just a check. You understand inlining a method inside the class declaration makes it implicitly `inline`?

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,89 +16,94 @@
...
 
-ThreadPool::ThreadPool(std::size_t workerCount): workerCount_(workerCount)
+private:
+    ThreadPool& threadPool_;
+};
+
+ThreadPool::ThreadPool(std::size_t workerCount)
+    : workerCount_(workerCount)
 {
     if (!workerCount_) {
```
rasendubi, line 27: Prefer `workerCount_ == 0`.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -114,32 +119,26 @@ void ThreadPool::shutdownNow()
...
 {
     for (std::size_t i = 0; i < workerCount_; ++i) {
-        workers_.push_back(std::thread(Worker(*this)));
+        workers_.emplace_back(Worker(*this));
     }
+
+    state_ = State::RUNNING;
 }
 
 void ThreadPool::stop(bool force)
```
rasendubi: Consider making two distinct functions: `stop` and `force_stop` as they are different cases. And second, seeing `stop(false)` is non-intuitive.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,89 +16,94 @@
...
+        }
 
-    KAA_UNLOCK(tasksLock);
+        tasks_.push_back(task);
+    }
 
     onNewTask_.notify_one();
 }
 
 void ThreadPool::awaitTermination(std::size_t seconds)
```
rasendubi, line 60: Any docs? `awaitTermination` must be called after `stop`.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,89 +16,94 @@
...
 {
-    KAA_MUTEX_UNIQUE_DECLARE(tasksLock, threadPoolGuard_);
-
-    if (isRun_) {
-        shutdownTimer_.reset(new KaaTimer<void()>("Thread pool shutdown timer"));
-        shutdownTimer_->start(seconds, [this] () { stop(true); } );
-    }
+    std::this_thread::sleep_for(std::chrono::seconds(seconds));
```
rasendubi: Non-conditionally sleeping is baaad.

client/client-multi/client-cpp/kaa/utils/ThreadPool.hpp
```diff
@@ -49,9 +47,15 @@ class ThreadPool : public IThreadPool {
     void start();
     void stop(bool force);
 
+    enum class State {
+        CREATED,
+        RUNNING,
+        PENDING_SHUTDOWN,
+        STOPPED
```
rasendubi: Add trailing comma.

client/client-multi/client-cpp/test/impl/utils/ThreadPoolTest.cpp
```diff
@@ -208,40 +208,45 @@ BOOST_AUTO_TEST_CASE(AwaitTerminationTest)
...
-        onStart.notify_all();
+        onStartCondition.notify_all();
     }
 
     std::size_t timeToWaitTwoTask = 2 * timeToWait - 1; // '-1' to wake up a bit more earlier than third task will start.
+
     threadPool.awaitTermination(timeToWaitTwoTask);
 
-    std::this_thread::sleep_for(std::chrono::seconds(timeToWaitAllTasks));
+    std::this_thread::sleep_for(std::chrono::seconds(timeToWaitAllTasks - timeToWaitTwoTask));
```
rasendubi: Some sleep magic is going on here.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -114,32 +119,26 @@ void ThreadPool::shutdownNow()
 void ThreadPool::start()
 {
     for (std::size_t i = 0; i < workerCount_; ++i) {
-        workers_.push_back(std::thread(Worker(*this)));
+        workers_.emplace_back(Worker(*this));
     }
+
+    state_ = State::RUNNING;
```
rasendubi, line 133: So...

``` c
pool.start();
pool.start();
```

... just doubles a number of workers... nice

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,89 +16,94 @@
...
     }
-}
 
-ThreadPool::ThreadPool(std::size_t workerCount): workerCount_(workerCount)
+private:
+    ThreadPool& threadPool_;
+};
+
+ThreadPool::ThreadPool(std::size_t workerCount)
+    : workerCount_(workerCount)
```
rasendubi, line 25: Why don't we just create workers in the constructor? That would simplify the class a lot.

client/client-multi/client-cpp/impl/context/AbstractExecutorContext.cpp
ghost, line 17: Yes, it is needed to trace locking/unlocking of mutex.
See KAA_MUTEX_LOCKING("useCountGuard_");

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost: Yes, but regarding to the changes, I just get rid of the separate definition of the operator() without any assumptions on performance improvements.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost, line 25: It is a some kind of a lazy initialization, to be exact, initialization on demand. 
Unless at least one task is added, no workers will start.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost, line 27: Agree what it is more obvious, but suggest to change it in scope of the "code style" task.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost, line 60: I will add documentation to IThreadPool.hpp

ghost, line 60: The implementation of this class is almost similar to Java ExecutorService
(see https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html)

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost: It is OK, this API call must be blocking.
The implementation of this class is almost similar to Java ExecutorService
(see https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html)

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost, line 133: It starts only once. There is check in add()
`if (state_ == State::CREATED) {
            start();
 }
`

client/client-multi/client-cpp/kaa/utils/ThreadPool.hpp
ghost: Why?

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost: OK

vadimol:
Looks good for me

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
rasendubi: I'm not against blocking; I'm against sleeping.

client/client-multi/client-cpp/kaa/utils/ThreadPool.hpp
rasendubi: General code style. It will be much easier to extend and shuffle values in the enum later.

client/client-multi/client-cpp/impl/context/AbstractExecutorContext.cpp
rasendubi, line 17: That means the file didn't compile before?

client/client-multi/client-cpp/kaa/utils/IThreadPool.hpp
```diff
@@ -27,11 +27,34 @@ typedef std::function<void()> ThreadPoolTask;
...
+     *
+     * @param task The task to add.
+     */
     virtual void add(const ThreadPoolTask& task) = 0;
 
+    /**
+     * @brief Blocks until all tasks have completed execution after
+     * a shutdown request, or the timeout occurs.
+     *
+     * @note Call only after @link shutdown() @endlink.
```
rasendubi: doxygen has auto-link property, thus `@link` and `@endlink` are not needed.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
rasendubi, line 60: The documentation at `IThreadPool` doesn't cover quirks of this particular implementation.

client/client-multi/client-cpp/kaa/utils/ThreadPool.hpp
ghost: Agree.

client/client-multi/client-cpp/kaa/utils/IThreadPool.hpp
ghost: Will edit.

forGGe:
Looks good after discussion

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
ghost: I will add a conditional exit on either timeout or all tasks complete

client/client-multi/client-cpp/impl/context/AbstractExecutorContext.cpp
ghost, line 17: Well, I rechecked the code and this snippet is unnecessary here. Thanks

ghost, line 17: Sorry, for the mess, but my code doesn't compile with this header, so I leave it.

dyosick: Sorry, for the mess, but my code doesn't compile with this header, so I leave it.

client/client-multi/client-cpp/impl/utils/ThreadPool.cpp
```diff
@@ -16,128 +16,150 @@
...
 void ThreadPool::awaitTermination(std::size_t seconds)
 {
-    KAA_MUTEX_UNIQUE_DECLARE(tasksLock, threadPoolGuard_);
+    {
+        KAA_MUTEX_UNIQUE_DECLARE(waitLock, threadPoolGuard_);
 
-    if (isRun_) {
-        shutdownTimer_.reset(new KaaTimer<void()>("Thread pool shutdown timer"));
-        shutdownTimer_->start(seconds, [this] () { stop(true); } );
+        if (state_ != State::PENDING_SHUTDOWN) {
```
rasendubi, line 65: Maybe it's better to call `shutdown()` here. (Or at least specify `@throw` in the doxygen comment.)

ghost, line 65: Added the `@throw` part

client/client-multi/client-cpp/kaa/utils/IThreadPool.hpp
```diff
@@ -27,11 +27,39 @@ typedef std::function<void()> ThreadPoolTask;
...
 
+    /**
+     * @brief Blocks until all tasks have completed execution after
+     * a shutdown request, or the timeout occurs.
+     *
+     * @note Call only after shutdown().
+     *
+     * @param seconds The maximum time in seconds to wait.
+     *
+     * * @throw std::logic_error The method was called without shutdown().
```
rasendubi, line 49: Too many stars

PR #892, opened 2016-07-20
KAA-1145: Quick fix for message of logging and name of method 'editNotificationSchema'.

v-chyzhevskyi:
Quick fix for message of logging and name of method 'editNotificationSchema'.


ashvayka:
LGTM

PR #897, opened 2016-07-21
Fix FindSqlite3.cmake

rasendubi:
Remove a section from FindSqlite3.cmake to search in a custom location.
There are a couple of reasons for that. The first one is that cmake already
provides means to specify additional search paths. The second is an
issue of priorities -- they were wrong, so cmake ended up selecting host
sqlite3 when I tried to cross-compile.

### DIFF
--- a/client/client-multi/client-cpp/Modules/FindSqlite3.cmake
+++ b/client/client-multi/client-cpp/Modules/FindSqlite3.cmake
@@ -11,14 +11,6 @@
 #    SQLITE3_INCLUDE_DIR
 #    SQLITE3_LIBRARY
 
-
-# FIND_PATH and FIND_LIBRARY normally search standard locations
-# before the specified paths. To search non-standard paths first,
-# FIND_* is invoked first with specified paths and NO_DEFAULT_PATH
-# and then again with no specified paths to search the default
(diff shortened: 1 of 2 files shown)


client/client-multi/client-cpp/Modules/FindSqlite3.cmake
```diff
@@ -37,31 +29,8 @@ IF (APPLE)
   ENDIF ()
 ENDIF (APPLE)
 
-IF ($ENV{SQLITE_ROOT})
```
ghost, line 40: Such changes will affect the process of building Kaa SDK, thus need to be discussed.

rasendubi, line 40: I don't think this "feature" was documented at all, so the impact on the end user is minimal.

akokhanovskyi:
What's the ticket number for this PR?

rasendubi:
@akokhanovskyi there is no.

@vadimol could you please open one?

vadimol:
The Jira is http://jira.kaaproject.org/browse/KAA-1319 .

Looks good for me

forGGe:
looks good

PR #903, opened 2016-07-25
KAA-1204 Sanity check in profile editing

nocs00:


ashvayka:
This was already done in https://github.com/kaaproject/kaa/pull/923

PR #919, opened 2016-07-27
KAA-1161 Big endian support for test_profile

balldir:
Line

> network_order_32 = KAA_HTONS(TEST_PUB_KEY_SIZE) << 16 | PUB_KEY_VALUE;

is not endian independent. It actually forms 2 fields that should be placed one after another. In this line, this fields placed properly only on the little-endian machine. 
/cc @forGGe @rasendubi


vadimol:
looks good for me

forGGe:
looks good

PR #923, opened 2016-07-27
KAA-1199 remove reduntant tempPassword field from users profile

nocs00:
This change also cover changes in
https://github.com/kaaproject/kaa/pull/903

So by merging this PR - 903 could be closed


ashvayka:
LGTM

PR #925, opened 2016-07-27
KAA-1305: ESP8266 C SDK requires g++5 package

tkhalymon:
Added g++5 installation to "Installing dependencies" label

### DIFF
--- a/doc/Programming-guide/Using-Kaa-endpoint-SDKs/C/SDK-ESP8266/index.md
+++ b/doc/Programming-guide/Using-Kaa-endpoint-SDKs/C/SDK-ESP8266/index.md
@@ -71,7 +71,7 @@ The detailed installation instructions can be found below.
 
 1. Prerequisites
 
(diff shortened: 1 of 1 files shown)


tkhalymon:
Output on clear environment:

sudo apt-get install autoconf libtool libtool-bin bison build-essential gawk git gperf flex texinfo libtool libncurses5-dev libc6-dev-amd64 python-serial libexpat-dev python-setuptools
Reading package lists... Done
Building dependency tree  
Reading state information... Done
Note, selecting 'libexpat1-dev' instead of 'libexpat-dev'
build-essential is already the newest version (12.1ubuntu2).
build-essential set to manually installed.
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 build-essential : Depends: gcc (>= 4:5.2) but it is not going to be installed
                   Depends: g++ (>= 4:5.2) but it is not going to be installed
 libtool : Depends: cpp but it is not going to be installed
E: Unable to correct problems, you have held broken packages.

vadimol:
It looks like a local problem with apt-get. The "build-essential" is enough.

rasendubi:
Closing because @vadimol comment.

PR #938, opened 2016-08-03
Kaa 1312: Fix problems with Configuration CTL Schema

Kirill380:


server/node/src/main/java/org/kaaproject/kaa/server/operations/service/cache/DeltaCacheKey.java
```diff
@@ -170,6 +142,8 @@ public String toString() {
         builder.append(endpointConfHash);
         builder.append(", resyncOnly=");
         builder.append(resyncOnly);
+        builder.append(", useRawSchema=");
```
ghost, line 145: ", useConfigurationRawSchema=" instead

PR #950, opened 2016-08-08
KAA-1438: [Android] Notification Sample Sdk Bug

maxml:
http://jira.kaaproject.org/browse/KAA-1438


client/client-multi/client-java-core/src/main/java/org/kaaproject/kaa/client/notification/DefaultNotificationManager.java
```diff
@@ -131,6 +130,12 @@ public void removeTopicListListener(NotificationTopicListListener listener) {
         List<Topic> topicList = new LinkedList<Topic>();
 
         synchronized (topics) {
+            for (Topic t : state.getTopics()) {
```
ashvayka, line 133: Same code is present in the constructor of this class. Maybe the root cause of the issue is that behaviour of "state.getTopics()" changes after start of the client?

PR #985, opened 2016-08-17
KAA-1144: fixed method for generating of record structure library.

v-chyzhevskyi:
Fixed method for generating of record structure library.


ashvayka:
LGTM

PR #993, opened 2016-08-18
KAA-1205: Unable to pass "Credential service" when creating application via REST API.

v-chyzhevskyi:
Fixed documentation for method of creat/edit application.


ashvayka:
LGTM

PR #995, opened 2016-08-18
KAA-1252 Pin Nixpks version

rasendubi:
/cc @forGGe @sparadiuk

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -27,10 +27,19 @@ let
 in
 
 { pkgs ? nixpkgs-16_03
+, pkgs-tools ? nixpkgs-bootstrap
 }:
 
 let
-  callPackage = pkgs.lib.callPackageWith (pkgs // self);
+  # We want to use latest versions of tools we have available
+  # (more checks, less false positives)
+  tools = {
+    doxygen = pkgs-tools.doxygen;
+    valgrind = pkgs-tools.valgrind;
+    cppcheck = pkgs-tools.cppcheck;
+  };
+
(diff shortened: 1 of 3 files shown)


forGGe:
lgtm

nix/default.nix
```diff
@@ -14,14 +14,36 @@
...
 
-{ system ? builtins.currentSystem
-, pkgs ? import <nixpkgs> { inherit system; }
+let
+  nixpkgs-bootstrap = import <nixpkgs> { };
+
+  nixpkgs-16_03 = import (nixpkgs-bootstrap.fetchFromGitHub {
+    owner = "NixOS";
+    repo = "nixpkgs-channels";
+    rev = "baf46b99e33005348fdbd083366c330be4b373f3";
```
vadimol, line 23: Is there any possibility to set a pointer to the release branch? It looks like that `rev` points to a specific commit in the development branch. For me, It is unclear, why this comment has been chosen.

rasendubi: Branches are moving things, which means `sha256` will be invalid pretty soon (which would break the whole Nix expression). The `rev` was chosen to point to the last commit in the `nixpkgs-channels/nixos-16.03` at the time of writing.

vadimol: Got it, thanks. I thought they have some sort of "fixed" release branches or tagged commits.

vadimol:
looks good for me

nix/default.nix
rasendubi: They don't all releases potentially can get security fixes, etc.

PR #1013, opened 2016-08-23
KAA-1373: Implement successful connection establishment detector

ohamov:


client/client-multi/client-cpp/kaa/NetworkAddrChangeDetector/NetworkAddrChangeDetector.hpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace kaa {
+
+/**
+ * The class is responsible for observation ip address changes and notifying of the subscriber
+ * if the ip address has been changed.
+ */
+class NetworkAddrChangeDetector {
+
+    std::string currentIp_;
```
RostakaGmfun: Let's move this to `private` section below.

client/client-multi/client-cpp/kaa/KaaClientContext.hpp
```diff
@@ -32,21 +32,23 @@ namespace kaa {
...
-    KaaClientContext(KaaClientProperties &properties, ILogger &logger,
-                     IExecutorContext &executorContext, IKaaClientStateStoragePtr state = nullptr)
-	  : properties_(properties), logger_(logger), executorContext_(executorContext), state_(state) {}
-
-    virtual KaaClientProperties              &getProperties() { return properties_; }
-    virtual ILogger                              &getLogger() { return logger_; }
-    virtual IKaaClientStateStorage               &getStatus() { return *state_; }
-    virtual IExecutorContext            &getExecutorContext() { return executorContext_; }
+    KaaClientContext(KaaClientProperties &properties, ILogger &logger, IExecutorContext &executorContext,
+        NetworkAddrChangeDetector& detector, IKaaClientStateStoragePtr state = nullptr)
```
RostakaGmfun: There are so many changes in tests due to new argument in constructor. How about defaulting it?

ohamov: Making it default implies necessity of the nullptr check each time before the detector is dereferenced.

ohamov:
@rasendubi

client/client-multi/client-cpp/kaa/IKaaClientContext.hpp
```diff
@@ -17,26 +17,27 @@
...
 public:
-    virtual KaaClientProperties              &getProperties() = 0;
-    virtual ILogger                              &getLogger() = 0;
-    virtual IKaaClientStateStorage               &getStatus() = 0;
-    virtual IExecutorContext            &getExecutorContext() = 0;
+    virtual KaaClientProperties                  &getProperties() = 0;
+    virtual ILogger                                  &getLogger() = 0;
+    virtual IKaaClientStateStorage                   &getStatus() = 0;
+    virtual IExecutorContext                &getExecutorContext() = 0;
+    virtual NetworkAddrChangeDetector    &getAddrChangeDetector() = 0;
```
rasendubi: Remove stupid alignment

client/client-multi/client-cpp/kaa/IKaaClientPlatformContext.hpp
```diff
@@ -56,6 +57,19 @@ class IKaaClientPlatformContext {
      */
     virtual IExecutorContext& getExecutorContext() = 0;
 
+    /**
+     * Sets subscriber which will be notified about endpoint ip address changes.
+     *
+     * @param[in] subscriber - The callback which will be called as soon as network address change has been detected.
```
rasendubi: Remove `-` -- it's not needed

client/client-multi/client-cpp/kaa/KaaClientContext.hpp
```diff
@@ -32,21 +32,23 @@ namespace kaa {
...
+        : properties_(properties), logger_(logger), executorContext_(executorContext), state_(state), detector_(detector) {}
+
+    virtual KaaClientProperties                  &getProperties() { return properties_; }
+    virtual ILogger                                  &getLogger() { return logger_; }
+    virtual IKaaClientStateStorage                   &getStatus() { return *state_; }
+    virtual IExecutorContext                &getExecutorContext() { return executorContext_; }
+    virtual NetworkAddrChangeDetector    &getAddrChangeDetector() { return detector_; }
```
rasendubi: Please, remove such alignment whenever you face it

client/client-multi/client-cpp/kaa/NetworkAddrChangeDetector/NetworkAddrChangeDetector.hpp
```diff
@@ -0,0 +1,69 @@
...
+   /**
+    * Constructs NetworkAddrDetector object.
+    *
+    * @param[in] subscriber The callback which is going to be called as soon as ip address change has been detected.
+    */
+    NetworkAddrChangeDetector(Subscriber subscriber) : subscriber_(subscriber) {}
+
+    /**
+     * Checks whether ip address has been changed.
+     * If it is the case - notifies the subscriber.
```
rasendubi: `s/ -/,/`

client/client-multi/client-cpp/impl/KaaClient.cpp
```diff
@@ -46,10 +46,9 @@ namespace kaa {
 
 KaaClient::KaaClient(IKaaClientPlatformContextPtr context, IKaaClientStateListenerPtr listener)
     : logger_(new DefaultLogger(context->getProperties().getClientId(), context->getProperties().getLogFileName())),
-      context_(context->getProperties(), *logger_, context->getExecutorContext()),
-      status_(new ClientStatus(context_)),
-      stateListener_(listener),
-      platformContext_(context)
+      detector_(context->getNetworkAddrChangeSubscriber()),
+      context_(context->getProperties(), *logger_, context->getExecutorContext(), detector_),
+      status_(new ClientStatus(context_)), stateListener_(listener), platformContext_(context)
```
rasendubi: One member initialization per line

client/client-multi/client-cpp/kaa/NetworkAddrChangeDetector/NetworkAddrChangeDetector.hpp
```diff
@@ -0,0 +1,69 @@
...
+#include <stdexcept>
+#include <functional>
+
+namespace kaa {
+
+/**
+ * The class is responsible for observation ip address changes and notifying of the subscriber
+ * if the ip address has been changed.
+ */
+class NetworkAddrChangeDetector {
```
rasendubi: This class does nothing. I mean... literally _nothing_.

I would just store a `std::function<void(const std::string &newAddrName)> on_reconnect;` in the client context (with default doing nothing) and call it whenever client reconnects.

The logic in `check()` is too simple to deserve a class. Furthermore, not checking `currentIp != ip` allows more usage by the user (e.g., check if we reconnect at all).

ohamov: The class _does_ the check. It is important because It is wrong to notify user if  IP address hasn't been changed.

> Furthermore, not checking currentIp != ip allows more usage by the user (e.g., check if we reconnect at all).

I see no benefits. The feature is not about notifying the user about reconnections, it is about IP address changes notifications. There is _failover feature_ which is responsible for _reconnections notifications_. 

On my opinion the check should be performed. Probably introducing a separate class for such simple logic isn't the best solution, but the check should be somewhere and moving it to the _KaaClientContext_ isn't the best solution too cuz I see no connection between client context and ip address change detection feature which itself related to the Client (Endpoint). Let's discuss it.

client/client-multi/client-cpp/kaa/IKaaClientContext.hpp
```diff
@@ -17,26 +17,27 @@
...
 public:
-    virtual KaaClientProperties              &getProperties() = 0;
-    virtual ILogger                              &getLogger() = 0;
-    virtual IKaaClientStateStorage               &getStatus() = 0;
-    virtual IExecutorContext            &getExecutorContext() = 0;
+    virtual KaaClientProperties        &getProperties() = 0;
+    virtual ILogger                    &getLogger() = 0;
+    virtual IKaaClientStateStorage     &getStatus() = 0;
+    virtual IExecutorContext           &getExecutorContext() = 0;
+    virtual IKaaClientStateListenerPtr getClientStateListener() = 0;
```
rasendubi: `&` is not part of names. (align on names.)

client/client-multi/client-cpp/kaa/KaaClientContext.hpp
```diff
@@ -26,27 +26,32 @@
...
+          logger_(logger),
+          executorContext_(executorContext),
+          state_(state),
+          stateListener_(stateListener) {}
+
+    virtual KaaClientProperties          &getProperties() { return properties_; }
+    virtual ILogger                      &getLogger() { return logger_; }
+    virtual IKaaClientStateStorage       &getStatus() { return *state_; }
+    virtual IExecutorContext             &getExecutorContext() { return executorContext_; }
+    virtual IKaaClientStateListenerPtr   getClientStateListener() { return stateListener_; }
```
rasendubi: Same here (alignment)

client/client-multi/client-cpp/kaa/KaaClientContext.hpp
```diff
@@ -26,27 +26,32 @@
...
 
 private:
     KaaClientProperties        &properties_;
-    ILogger                        &logger_;
-    IExecutorContext      &executorContext_;
-    IKaaClientStateStoragePtr        state_;
+    ILogger                    &logger_;
+    IExecutorContext           &executorContext_;
+    IKaaClientStateStoragePtr  state_;
+    IKaaClientStateListenerPtr stateListener_;
```
rasendubi: and here

client/client-multi/client-cpp/kaa/KaaClientContext.hpp
```diff
@@ -26,27 +26,32 @@
...
-    KaaClientContext(KaaClientProperties &properties, ILogger &logger,
-                     IExecutorContext &executorContext, IKaaClientStateStoragePtr state = nullptr)
-	  : properties_(properties), logger_(logger), executorContext_(executorContext), state_(state) {}
-
-    virtual KaaClientProperties              &getProperties() { return properties_; }
-    virtual ILogger                              &getLogger() { return logger_; }
-    virtual IKaaClientStateStorage               &getStatus() { return *state_; }
-    virtual IExecutorContext            &getExecutorContext() { return executorContext_; }
+    KaaClientContext(KaaClientProperties &properties, ILogger &logger, IExecutorContext &executorContext,
+        IKaaClientStateListenerPtr stateListener, IKaaClientStateStoragePtr state = nullptr)
```
rasendubi: Can we add default value for stateListener and move in the end, so we won't change all the tests?

rasendubi:

rasendubi:
LGTM

client/client-multi/client-cpp/impl/channel/impl/DefaultOperationTcpChannel.cpp
```diff
@@ -252,6 +252,10 @@ void DefaultOperationTcpChannel::openConnection()
         return;
     }
 
+    /* Check whether ip has been changed */
```
ghost: Strange comment. As I understand from the line below, you don't check but noify about a local address.

client/client-multi/client-cpp/impl/channel/impl/DefaultOperationTcpChannel.cpp
```diff
@@ -252,6 +252,10 @@ void DefaultOperationTcpChannel::openConnection()
...
+    context_.getClientStateListener()->onConnectionEstablished(sock_->local_endpoint().address().to_string(),
```
ghost: I think it is better to notify the channel manager about new connection and the latter will pass this info to the listener or to some other modules. In other words to use the same approach like in the case when connection is failed.
Besides I suggest to wrap both parameters in a single structure like IPConnectionInfo. Using that approach lets us simply add a field or two if we want to specify destination's host/port.

client/client-multi/client-cpp/kaa/IKaaClientContext.hpp
```diff
@@ -17,26 +17,27 @@
...
 public:
-    virtual KaaClientProperties              &getProperties() = 0;
-    virtual ILogger                              &getLogger() = 0;
-    virtual IKaaClientStateStorage               &getStatus() = 0;
-    virtual IExecutorContext            &getExecutorContext() = 0;
+    virtual KaaClientProperties       &getProperties() = 0;
+    virtual ILogger                   &getLogger() = 0;
+    virtual IKaaClientStateStorage    &getStatus() = 0;
+    virtual IExecutorContext          &getExecutorContext() = 0;
+    virtual IKaaClientStateListenerPtr getClientStateListener() = 0;
```
ghost: Could we return a reference?

client/client-multi/client-cpp/kaa/IKaaClientStateListener.hpp
```diff
@@ -36,58 +36,68 @@ class IKaaClientStateListener {
...
+
+    /**
+     *
+     * On connection established. The method is called as soon as the connection with either bootstrap
+     * or operation server is established.
+     *
+     * @param[in] endpointIp The internet address of endpoint which is currently in use.
+     * @param[in] serverIp   The server ip address the endpoint is connected to.
+     */
+    virtual void onConnectionEstablished(const std::string& endpointIp, const std::string& serverIp) {}
```
ghost: See the comment about wrapping these parameters in a single structure.

client/client-multi/client-cpp/kaa/KaaClientContext.hpp
```diff
@@ -26,27 +26,33 @@
...
 {
 public:
-    KaaClientContext(KaaClientProperties &properties, ILogger &logger,
-                     IExecutorContext &executorContext, IKaaClientStateStoragePtr state = nullptr)
-	  : properties_(properties), logger_(logger), executorContext_(executorContext), state_(state) {}
-
-    virtual KaaClientProperties              &getProperties() { return properties_; }
-    virtual ILogger                              &getLogger() { return logger_; }
-    virtual IKaaClientStateStorage               &getStatus() { return *state_; }
+    KaaClientContext(KaaClientProperties &properties, ILogger &logger, IExecutorContext &executorContext,
```
ghost: Align constuctor's parameters like you did for initialization list

client/client-multi/client-cpp/impl/KaaClient.cpp
```diff
@@ -38,21 +38,19 @@
...
+      context_(platformContext->getProperties(), *logger_, platformContext->getExecutorContext(), nullptr, listener),
       status_(new ClientStatus(context_)),
-      stateListener_(listener),
-      platformContext_(context)
+      platformContext_(platformContext)
 {
-    if (!stateListener_) {
-        stateListener_ = std::make_shared<DummyKaaClientStateListener>();
+    if (context_.getClientStateListener() == nullptr) {
+        throw KaaException("State listener is null");
```
ghost: Listener is not mandatory! I used a dummy implementation just to not to check for null in other places.

client/client-multi/client-cpp/impl/http/HttpClient.cpp
```diff
@@ -66,6 +66,10 @@ std::shared_ptr<IHttpResponse> HttpClient::sendRequest(const IHttpRequest& reque
     sock_.connect(ep, errorCode);
     checkError(errorCode);
 
+    /* Check whether ip has been changed */
+    context_.getClientStateListener()->onConnectionEstablished(sock_.local_endpoint().address().to_string(),
```
ghost: See the comment above.

client/client-multi/client-cpp/impl/KaaClient.cpp
rasendubi: We're still using dummy implementation as default argument

ghost: Why did you decide to transform an abstract class to one that can be instantiated and leave "I" in the name of this class? I mean IKaaClientStateListener

rasendubi: Because "backward compatibility". We'll drop `I` in future releases.

ghost: Is there a ticket for this?

client/client-multi/client-cpp/kaa/IKaaClientContext.hpp
```diff
@@ -17,26 +17,27 @@
...
-#include <memory>
 #include <kaa/IKaaClientStateStorage.hpp>
+#include <kaa/KaaClientStateListener.hpp>
 
 namespace kaa {
 
 class KaaClientProperties;
 class ILogger;
 class IExecutorContext;
+class NetworkAddrChangeDetector;
```
ghost: What is this?

client/client-multi/client-cpp/kaa/channel/IKaaChannelManager.hpp
```diff
@@ -18,13 +18,27 @@
...
 #include "kaa/channel/IDataChannel.hpp"
 #include "kaa/channel/ITransportConnectionInfo.hpp"
 #include "kaa/channel/connectivity/IConnectivityChecker.hpp"
 
 namespace kaa {
 
+/**
+ * The structure represents connection information in readable format.
+ */
+struct EpConnectionInfo {
```
ghost: How about IpConnectionInfo?

client/client-multi/client-cpp/test/headers/channel/MockChannelManager.hpp
```diff
@@ -59,6 +59,7 @@ class MockChannelManager: public IKaaChannelManager {
 
     virtual void setConnectivityChecker(ConnectivityCheckerPtr checker) { ++onSetConnectivityChecker_; }
 
+    virtual void onConnected(const EpConnectionInfo& connection)  override {}
```
ghost: Could you add "override" for all methods or don't use at all?)

client/client-multi/client-cpp/kaa/KaaClientStateListener.hpp
```diff
@@ -29,66 +31,76 @@ class KaaException;
...
+    /**
+     *
+     * On connection established. The method is called as soon as the connection with either bootstrap
+     * or operation server is established.
+     *
+     * @param[in] endpointIp The internet address of endpoint which is currently in use.
+     * @param[in] serverIp   The server ip address the endpoint is connected to.
      */
-    virtual void onStopFailure(const KaaException& exception) = 0;
+    virtual void onConnectionEstablished(const EpConnectionInfo& connection) {}
```
ghost: Could you open a story to add a callback for a connection failure?

client/client-multi/client-cpp/kaa/channel/IKaaChannelManager.hpp
ohamov: I'd prefer the current one. If there is a need to add some non-ip-related data (for example the type of the server), this name will be better fit.

rasendubi: I'm in favor of `ConnectionInfo`

client/client-multi/client-cpp/kaa/KaaClientStateListener.hpp
ohamov: Shell I? There is the `failover strategy` for this purpose, though It covers more than just _connections failures_. Maybe It is more suitable to expand failover strategy reasons?

client/client-multi/client-cpp/kaa/channel/IKaaChannelManager.hpp
rasendubi: Discussed. Expand `Ep` to `Endpoint` and I'm fine (we're not saving keystrokes here)

client/client-multi/client-cpp/kaa/KaaClientStateListener.hpp
```diff
@@ -29,66 +31,76 @@ class KaaException;
...
+     */
+    virtual void onStopFailure(const KaaException& exception) {}
+
+    /**
+     *
+     * On connection established. The method is called as soon as the connection with either bootstrap
+     * or operation server is established.
+     *
+     * @param[in] endpointIp The internet address of endpoint which is currently in use.
+     * @param[in] serverIp   The server ip address the endpoint is connected to.
```
rasendubi: doxygen is now outdated

client/client-multi/client-cpp/kaa/channel/IKaaChannelManager.hpp
```diff
@@ -18,13 +18,27 @@
...
 namespace kaa {
 
+/**
+ * The structure represents connection information in readable format.
+ */
+struct EndpointConnectionInfo {
+    std::string endpointIp_;
+    std::string serverIp_;
+
+    EndpointConnectionInfo() = default;
```
rasendubi: Do we need default constructor? I would remove

client/client-multi/client-cpp/kaa/channel/IKaaChannelManager.hpp
```diff
@@ -18,13 +18,27 @@
...
 #include "kaa/channel/ITransportConnectionInfo.hpp"
 #include "kaa/channel/connectivity/IConnectivityChecker.hpp"
 
 namespace kaa {
 
+/**
+ * The structure represents connection information in readable format.
+ */
+struct EndpointConnectionInfo {
+    std::string endpointIp_;
```
ghost: Could you please add a server type?

client/client-multi/client-cpp/kaa/KaaClientStateListener.hpp
```diff
@@ -14,11 +14,13 @@
...
  */
 
-#ifndef IKAACLIENTSTATELISTENER_HPP_
-#define IKAACLIENTSTATELISTENER_HPP_
+#ifndef KAACLIENTSTATELISTENER_HPP_
+#define KAACLIENTSTATELISTENER_HPP_
 
 #include <memory>
 
+#include "kaa/channel/IKaaChannelManager.hpp"
```
ghost: Move EndpointConnectionInfo to a separate header in order to not to include IKaaChannelManager.hpp

client/client-multi/client-cpp/kaa/channel/IKaaChannelManager.hpp
ohamov: I left it because of [this](https://github.com/kaaproject/kaa/pull/1013/files#diff-a3f7bfb8dbedef8645fe23657b827651R57).  Though I can replace it with `EndpointConnectionInfo foo("","");`, but this is not quite pretty.

ohamov:
@dyosick

ghost:
LGTM

PR #1020, opened 2016-08-26
KAA-1367: Failed build and linking C++ part on OS X

vasner:
Add "if else" to avoid MSVC compiler fails;

Reviewers: @RostakaGmfun and @dyosick


RostakaGmfun:
LGTM
/cc @vadimol

rasendubi:
LGTM

vadimol:
+

PR #1025, opened 2016-08-29
KAA-1352: [REST API] Incorrect behavior in method Create/Edit user.

v-chyzhevskyi:
Added description of 400 error for method create/edit user.


ikulikov:

PR #1031, opened 2016-08-31
KAA-1357: Sidebar patch

dr3amf4ll:
According to http://jira.kaaproject.org/browse/KAA-1357


Oldnickus:

PR #1054, opened 2016-09-09
KAA-1341. Fix problem with long names of event classes

Kirill380:


ikulikov:

PR #1071, opened 2016-09-14
KAA-1191: EndpointKeyHash field is not refreshed in Notificaion topic

Pyshankov:


vadimol:
@kaa-leeroy re-test this please.

PR #1096, opened 2016-09-22
 KAA-1423: [WIP] Added Endpoint Specific Configuration Support

agulenko:


kaa-leeroy:

PR #1101, opened 2016-09-22
KAA-1135: Using Kaa endpoint SDKs - reviewed.

hunch7:


doc/Programming-guide/Using-Kaa-endpoint-SDKs/index.md
```diff
@@ -11,14 +11,14 @@ sort_idx: 40
 * TOC
 {:toc}
 
-Developing an IoT solution we all the time facing the same routine: creating network communication stack, log delivery functionality, event exchange between endpoints, etc.
-The Kaa platform already provides all of this functionality, and you can get it out of the box using Kaa endpoint SDK.
+Developing a [Kaa instance]({{root_url}}Glossary/#kaa-instance-kaa-deployment) requires following the same routine: creating network communication stack, log delivery functionality, event exchange between endpoints, etc.
```
vadimol, line 14: It is unclear to me. I think you've missed the key point of this sentence.
The first version:
_Developing an IoT solution we all the time facing the same routine: creating network communication stack, log delivery functionality, event exchange between endpoints, etc._
means that the typical IoT solution requires implementing of the same routines.

The _Kaa instance_ already provides all of these routines and it does not require them.

vadimol (changes requested):
Reviewed. Please, take a look.

doc/Programming-guide/Using-Kaa-endpoint-SDKs/index.md
vadimol: Or, maybe, I just don't understand you correctly.

hunch7: I think you said the same but in other words, which means your understanding is correct anyway :) All I did there is I changed 'IoT solution' to 'Kaa instance' to be more precise why I mention that 'routine' -- I don't think it's safe to assume it's the same routine for _every_ IoT solution out there.

PR #1102, opened 2016-09-23
KAA-1442: REST method editUser do not send email with temp password

nocs00:


nocs00:
@kaa-leeroy retest this please.

nocs00:
@kaa-leeroy retest this please.

PR #1126, opened 2016-10-03
KAA-1356: Fix cross-compilation guide C++ SDK (Intel Edison)

GeorgePostolakiy:


doc/Programming-guide/Using-Kaa-endpoint-SDKs/C++/SDK-Edison/index.md
```diff
@@ -81,71 +84,108 @@ Refer to [the official Edison getting started guide](https://software.intel.com/
...
+1. Build Botan.
 
-        source /opt/poky-edison/1.6.1/environment-setup-core2-32-poky-linux
+        cd $EDISON_ROOT
+        source /opt/poky-edison/1.6/environment-setup-core2-32-poky-linux
         wget http://botan.randombit.net/releases/Botan-1.11.27.tgz
         tar -xvzf Botan-1.11.27.tgz
         cd Botan-1.11.27
-        python configure.py --cpu=x86_32 --cc-bin=${CROSS_COMPILE}g++ --prefix=${SDKTARGETSYSROOT}/usr
+        /usr/bin/python -E configure.py --cpu=x86_32 --cc-bin=${CROSS_COMPILE}g++ --prefix=${SDKTARGETSYSROOT}/usr
```
rasendubi: My system don't have `/usr` directory at all. Prefer user's `$PATH` over hard-coded paths.

mlysenko: Edison SDK environment modifies `$PATH`:

```
    export PATH=/opt/poky-edison/1.6.1/sysroots/x86_64-pokysdk-linux/usr/bin:/opt/poky-edison/1.6.1/sysroots/x86_64-pokysdk-linux/usr/bin/i586-poky-linux:$PATH
```

SDK puts its own path before system path. We use `/usr/bin/python` in order to run Python of the host machine. I suppose it will work for 99% of popular Linux distributions, including Ubuntu.

rasendubi: Sysroot should never end up in user's PATH because you never want (or can) to run these tools from a host.

There are many legitimate reasons for having `python` in a different location even on Ubuntu (e.g., `/usr/local/bin/python`), so I'm not buying that.

vadimol: As far as I understend, the issues here is that we need to use the "native" (host) `python` in the command.
The `source /opt/poky-edison/1.6/environment-setup-core2-32-poky-linux`
overwrites the default path for Python interpreter.
However, It also defines environment variables that are required to run the command (like SDKTARGETSYSROOT).

The quickest way to workaround the problem is:
1) `export HOST_PYTHON=$(which python)`
2) `source /opt/poky-edison/1.6/environment-setup-core2-32-poky-linux`
3) `$(HOST_PYTHON)  -E configure.py --cpu=x86_32 --cc-bin=${CROSS_COMPILE}g++ --prefix=${SDKTARGETSYSROOT}/usr`

vadimol:

rasendubi (approved):

vadimol:
Jenkins fails due to http://jira.kaaproject.org/browse/KAA-1432

vadimol (approved):

PR #1155, opened 2016-10-12
KAA-1486:  [C-SDK]  Fixed an error of the profile hash

sparadiuk:


client/client-multi/client-c/src/kaa/platform-impl/common/sha.c
```diff
@@ -24,7 +24,7 @@
 
 kaa_error_t ext_calculate_sha_hash(const char *data, size_t data_size, kaa_digest digest)
 {
-    if (!digest || !data || !data_size) {
+    if (!digest || (!data && data_size)) {
```
mlysenko: I think it's better to add one more condition `(data && !data_size)` :

```
    if (!digest || (data_size && !data) || (data && !data_size)) {
        return KAA_ERR_BADPARAM;
    }
```

In this case the function behavior won't change comparing to original function.  Passing the 0 size to function `ext_calculate_sha_hash(valid_data, 0, hash)` should result in error `KAA_ERR_BADPARAM`. Also we won't need to modify unit tests.

client/client-multi/client-c/test/test_kaa_common.c
```diff
@@ -39,7 +39,7 @@ void test_profile_update(void **state)
     err = ext_calculate_sha_hash(NULL, 50, calculated_hash);
     ASSERT_NOT_EQUAL(err, KAA_ERR_NONE);
     err = ext_calculate_sha_hash(body, 0, calculated_hash);
-    ASSERT_NOT_EQUAL(err, KAA_ERR_NONE);
+    ASSERT_EQUAL(err, KAA_ERR_NONE);
```
mlysenko, line 42: We shouldn't change the function behavior. Please look at my comment above.

client/client-multi/client-c/src/kaa/platform-impl/common/sha.c
vadimol: I agree with Maxim. 
Also, take a look at the _mbedtls_sha1_ function in mbedtls library. It does the same steps that are represented below. I'd recommend to use it.

Also, is it really necessary to call this function with NULL buffer?
I mean _ext_calculate_sha_hash(NULL, 0, profile_manager->profile_hash);_ in kaa_profile.c

What does it mean? Let me say, it is not the SHA-1 from zero.
Ideally,  I'd prefer to leave this function without changes and implement better logic into the caller function.

vadimol: And also, don't forget to change function documentation in the header file. All changes **_must**_ be reflected there.

vadimol (changes requested):
See comments

client/client-multi/client-c/src/kaa/platform-impl/common/sha.c
rasendubi: @mlysenko, your check is not needed. You can read 0 bytes from any pointer really. (Imagine statically-allocated array that has no data yet.)

client/client-multi/client-c/test/test_kaa_common.c
rasendubi: If the function behaviour is bogus, we should.

client/client-multi/client-c/src/kaa/platform-impl/common/sha.c
rasendubi: Expand according to MISRA:

``` c
if (digest == NULL || (data == NULL && data_size != 0)) {
```

rasendubi (commented):
I'm in favor of this change.

sha1 is defined for an empty byte array, so `data_size == 0` is a valid case.

`data == NULL && data_size == 0` is a common practice in C, so that's also fine.

client/client-multi/client-c/src/kaa/platform-impl/common/sha.c
sparadiuk: Changed to MISRA

vadimol (approved):

vadimol:
LGTM after local discussion

PR #1161, opened 2016-10-13
KSF-88: Getting started - fixed upon developer review.

hunch7:


Slash32 (changes requested):
I would not recommend dropping whole the TIP part with description of the Network configuration, modes, etc.

Could this be just extended with the information that the Sandbox is now capable of working out of the box (provided the Hostname is correct)?

The main advantage of Bridged is that the Sandbox can detect and apply the Kaa Host correctly. While NAT does not provide such support and is better in other aspects.

hunch7:
Fixed.

Slash32 (approved):

doc/Getting-started/index.md
```diff
@@ -53,24 +53,17 @@ The current version of Kaa Sandbox supports Oracle VirtualBox version 5.1.2 and
...
 For optimal performance, we recommend at least 4 GB of RAM and at least 2 CPUs.
+Apply your settings and start the virtual machine.
 
-	>**TIP:** Configure your network settings.  
-	>The default VM networking configuration is NAT, which enables your host machine to connect to the [Kaa server]({{root_url}}Glossary/#kaa-server).
-	>This means that you will need to deploy your [Kaa client]({{root_url}}Glossary/#kaa-client) applications to your host machine to enable their communication with the Sandbox server.
-	>
-	>You can deploy the Kaa client applications to any hardware other than your host machine, and access Kaa web interfaces from your network devices.
+	>**NOTE:** Your instance of Kaa Sandbox works out of the box with the default VM networking configuration set to NAT.
+	>This enables your host machine to connect to the [Kaa server]({{root_url}}Glossary/#kaa-server).
```
rasendubi: As far as I know, in current configuration this allows connecting from any machine, not only host.

doc/Getting-started/index.md
```diff
@@ -53,24 +53,17 @@ The current version of Kaa Sandbox supports Oracle VirtualBox version 5.1.2 and
...
+	>However, you may want to switch to the **Bridged Adapter** mode.
 	>To do that, open the **Oracle VM VirtualBox Manager**, click **Settings** > **Network** and change the **Attached to** option to **Bridged Adapter**.
-	>
-	>To configure the networking mode, switch to the **Network** tab in the VM settings.
-	>Click **Advanced** to unfold the drop-down menu, and make sure the **Promiscuous Mode** option is set to **Allow All**.
-	{: .tip}
+	>The advantage of using this mode is that your Kaa Sandbox will work correctly even if you change your local machine IP.
+	{:.note}
+5. Wait until the VirtualBox starts up and open [127.0.0.1:9080/sandbox](http://127.0.0.1:9080/sandbox) URL in your browser.
```
rasendubi, line 64: Does this work for bridged adapter?

hunch7: No, it doesn't. Actually, the Sandbox console displays a pretty clear message regarding what IP your Sandbox is on, for both NAT and Bridged Adapter, so our dear users won't get lost anyway.

doc/Getting-started/index.md
hunch7: Fixed.

doc/Getting-started/index.md
Slash32: Actually, Sandbox will work correctly provided it is in Bridged mode and gets rebooted after the IP change. Applications will not work until their SDKs are updated with the new IP address.

What is meant here is that for NAT mode, you will need to also change the Management->Kaa Host manually before downloading SDKs.

Slash32: Thus, basically, the text presented is correct. It is just not full. But better explanation might be too complex.

rasendubi (approved):

PR #1164, opened 2016-10-13
KAA-1481: 1-thread callback executor (Java client) causes issue with "nested" callback calls

Serega290696:


kaa-leeroy:

PR #1174, opened 2016-10-18
Add try_sync() for profile transport

ohamov:
Because of https://github.com/kaaproject/kaa/commit/0e5e524314205796fc3146d696e7042bd103c448, transport errors are not ignored anymore. It can somehow brake the SDK logic. So the patch introduces the function which tries to sync, but in case of failure the function doesn't reports the error.

### DIFF
--- a/client/client-multi/client-c/src/extensions/profile/kaa_profile.c
+++ b/client/client-multi/client-c/src/extensions/profile/kaa_profile.c
@@ -192,6 +192,10 @@ kaa_error_t kaa_profile_force_sync(kaa_profile_manager_t *self)
     return KAA_ERR_NONE;
 }
 
+static void kaa_profile_try_sync(kaa_profile_manager_t *self)
+{
+  (void)kaa_profile_force_sync(self);
+}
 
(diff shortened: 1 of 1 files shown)


client/client-multi/client-c/src/extensions/profile/kaa_profile.c
```diff
@@ -378,7 +382,7 @@ kaa_error_t kaa_profile_handle_server_sync(kaa_profile_manager_t *self
     if (extension_options & KAA_PROFILE_RESYNC_OPTION) {
         self->need_resync = true;
         KAA_LOG_INFO(self->logger, KAA_ERR_NONE, "Going to resync profile...");
-        error_code = kaa_profile_force_sync(self);
+        kaa_profile_try_sync(self);
```
rasendubi, line 385: Just use void-cast here and there and describe the reason in the comment. An additional function is not needed.

rasendubi (changes requested):

rasendubi:
I've reopened KAA-1403, so please add ticket number.

ohamov:
suppressed by https://github.com/kaaproject/kaa/pull/1175

PR #1175, opened 2016-10-19
KAA-1403 Ignore error codes

ohamov:
The functions can be called while the channels are not
initialized. It's permitted and the error codes should be
ignored in this case.

### DIFF
--- a/client/client-multi/client-c/src/extensions/profile/kaa_profile.c
+++ b/client/client-multi/client-c/src/extensions/profile/kaa_profile.c
@@ -192,7 +192,6 @@ kaa_error_t kaa_profile_force_sync(kaa_profile_manager_t *self)
     return KAA_ERR_NONE;
 }
 
-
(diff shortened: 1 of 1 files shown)


vadimol (approved):

client/client-multi/client-c/src/extensions/profile/kaa_profile.c
```diff
@@ -452,10 +453,8 @@ kaa_error_t kaa_profile_manager_set_endpoint_access_token(kaa_profile_manager_t
     kaa_error_t error =  kaa_status_set_endpoint_access_token(self->status, token);
     if (error == KAA_ERR_NONE) {
         self->need_resync = true;
-        /* The function can be called before the channels are created, thus
-         * ignore the error code.
-         */
-        kaa_profile_force_sync(self);
+        /* Ignoring an error code: the channels can be not initialized */
```
rasendubi, line 456: btw, previous comment was better

PR #1195, opened 2016-10-25
KAA-1512 [C SDK]Inconsistent naming of RSA key files.

ghost:
RSA keys filenames in C SDK change to be the same as in C++ and Java SDKs.

### DIFF
--- a/client/client-multi/client-c/src/kaa/platform-impl/common/key_utils.c
+++ b/client/client-multi/client-c/src/kaa/platform-impl/common/key_utils.c
@@ -34,8 +34,8 @@
 
 #include <kaa_rsa_key_gen.h>
 
-#define KAA_PRIVATE_KEY_STORAGE "private.key"
-#define KAA_PUBLIC_KEY_STORAGE "public.key"
+#define KAA_PRIVATE_KEY_STORAGE "key.private"
+#define KAA_PUBLIC_KEY_STORAGE "key.public"
 
 #define KAA_RSA_PUBLIC_KEY_LENGTH_MAX  294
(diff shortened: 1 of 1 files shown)


ohamov (changes requested):
Please change commit message according to the pattern:
`KAA-[TICKET NUMBER] Short changes brief`.
The same story is for the pull request name,

vadimol (approved):

client/client-multi/client-c/src/kaa/platform-impl/common/key_utils.c
```diff
@@ -34,8 +34,8 @@
 
 #include <kaa_rsa_key_gen.h>
 
-#define KAA_PRIVATE_KEY_STORAGE "private.key"
-#define KAA_PUBLIC_KEY_STORAGE "public.key"
+#define KAA_PRIVATE_KEY_STORAGE "key.private"
+#define KAA_PUBLIC_KEY_STORAGE "key.public"
```
rasendubi, line 38: SSH stores key pair as `key_name` for private and `key_name.pub` for public. It might be wise to use same naming scheme.

ohamov (approved):

client/client-multi/client-c/src/kaa/platform-impl/common/key_utils.c
vadimol: The Java SDK and C++ SDK uses these names. Let's accept this and open the Jira tickets to change names in next version.

rasendubi: Sounds good

PR #1203, opened 2016-10-26
KAA-1519: Removed duplicated SQLite installation step

vzhukovskyi:


kaa-leeroy:

PR #1204, opened 2016-10-26
KAA-1518: Added intro, set View instead of Supported

vzhukovskyi:


kaa-leeroy:

PR #1238, opened 2016-11-02
KAA-1510 Add Ubuntu 16.04 support

ohamov:


doc/Programming-guide/Using-Kaa-endpoint-SDKs/C++/SDK-Linux/index.md
```diff
@@ -27,7 +28,7 @@ Before building the C++ endpoint SDK, install the following components on your m
 1. Install g++, [CMake](https://cmake.org/download/), [Boost](http://www.boost.org/users/download/) and [SQLite3](https://sqlite.org/download.html):
 
    ```
-   sudo apt-get install g++ cmake libboost1.55-all-dev libsqlite3-0 libsqlite3-dev
+   sudo apt-get install g++ cmake python libboost-all-dev libsqlite3-0 libsqlite3-dev
```
Slash32: Does this change mean also that for Ubuntu 14.04 libboost-all-dev now refers to libboost1.55-all-dev (or above)?

Slash32 (changes requested):
I would rather make the commands separated by a tabs like it is done on other Documentation pages.

doc/Programming-guide/Using-Kaa-endpoint-SDKs/C++/SDK-Linux/index.md
ohamov: No, for Ubuntu 14.04 libboost-all-dev refers to libboost1.54-all-dev. But I'm convinced that [we can build cpp client using 1.54](https://github.com/kaaproject/kaa/blob/release-0.10/client/client-multi/client-cpp/CMakeLists.txt#L393) . Though, thx for the comment, I found that the patch also disables `avrogencpp` build. So, I'm going to change the guide: for Ubuntu 16.04 boost should be built from the sources or we should create another patch which disables the tests build for Avro, but doesn't disable avrogencpp build. @Slash32, what option do you think fits better?

vadimol: I wouldn't be so sure regarding boost version. Have you checked that C++ SDK works fine with 1.54?

doc/Programming-guide/Using-Kaa-endpoint-SDKs/C++/SDK-Linux/index.md
```diff
@@ -53,6 +86,12 @@ To install, proceed as follows:
    sudo ln -s /usr/local/include/botan-1.11/botan /usr/local/include/botan
    ```
 
+5. Make sure that dynamic loader will find the libraries in `/usr/local/lib`:
+
+   ```
+   export LD_LIBRARY_PATH=/usr/local/lib:${LD_LIBRARY_PATH}
```
Slash32: Why not to run ldconfig(8) for this?

ohamov: Is there any advantage of using `ldconfig` instead of `LD_LIBRARY_PATH`?

Slash32: Yes, ldconfig has system-wide effect including already running processes and environments.
And the 'export' way affects only the environment (process) this was done in and is dropped with its exit.

vadimol (approved):

rasendubi:
Manually installing to `/usr` is bad. It's controlled by the system package manager, so it may come as a surprise to it. (And all kind of bad things could happen.)

rasendubi:
@GamovCoder, please revert to the previous version and add `ldconfig` call along with a short explanation why it's needed. (It might be good to keep both versions if we explain the difference, but that's up to you.)

doc/Programming-guide/Using-Kaa-endpoint-SDKs/C++/SDK-Linux/index.md
```diff
@@ -50,7 +83,12 @@ To install, proceed as follows:
    cd botan-1.11.28/
    ./configure.py
    sudo make install
-   sudo ln -s /usr/local/include/botan-1.11/botan /usr/local/include/botan
+   sudo ln -s /usr/include/botan-1.11/botan /usr/include/botan
```
rasendubi: Please sync with #1244.

doc/Programming-guide/Using-Kaa-endpoint-SDKs/C++/SDK-Linux/index.md
```diff
@@ -50,7 +83,12 @@ To install, proceed as follows:
...
+   ```
+5. Update links to libraries:
```
rasendubi: Expand little on what `ldconfig` does.

doc/Programming-guide/Using-Kaa-endpoint-SDKs/C++/SDK-Linux/index.md
ohamov: removed the line, guess my request should merged after the one you mentioned

vadimol:
Merged related requests

rasendubi (approved):

rasendubi:
@GamovCoder, please, backport to master.

PR #1250, opened 2016-11-09
KAA-1573: [C++] No linking to threading library on Intel Edison platform

RostakaGmfun:
/cc @Slash32 @vadimol

### DIFF
--- a/client/client-multi/client-cpp/CMakeLists.txt
+++ b/client/client-multi/client-cpp/CMakeLists.txt
@@ -133,7 +133,7 @@
 #       Default: `0`.
 
 cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
-project(KaaCppSdk CXX)
+project(KaaCppSdk C CXX)
 
 macro(DECCHAR2HEX VAR VAL)
     if(${VAL} LESS 10)
@@ -393,6 +393,7 @@ endif()
 find_package(Boost 1.54 REQUIRED COMPONENTS log system thread)
 find_package(Avro REQUIRED)
 find_package(Botan REQUIRED)
+find_package(Threads REQUIRED)
 
 if(KAA_WITH_SQLITE_LOG_STORAGE)
     find_package(Sqlite3 REQUIRED)


Slash32 (approved):

client/client-multi/client-cpp/CMakeLists.txt
```diff
@@ -133,7 +133,7 @@
 #       Default: `0`.
 
 cmake_minimum_required(VERSION 2.8.12 FATAL_ERROR)
-project(KaaCppSdk CXX)
+project(KaaCppSdk C CXX)
```
vadimol, line 136: Is it necessary?

client/client-multi/client-cpp/CMakeLists.txt
```diff
@@ -393,6 +393,7 @@ endif()
 find_package(Boost 1.54 REQUIRED COMPONENTS log system thread)
 find_package(Avro REQUIRED)
 find_package(Botan REQUIRED)
+find_package(Threads REQUIRED)
```
vadimol, line 396: Have you investigated, why the
_find_package(Boost 1.54 REQUIRED COMPONENTS log system **thread**)_
cannot find this library on Intel Edison?

RostakaGmfun: Indeed, the `find_package(Boost 1.54 REQUIRED COMPONENTS log system thread)` successfully finds the **boost.threads** library, but don't confuse it with **Threads** package which provides platform-specific threading implementation. Initially it was expected that the above command will take care of finding the Threads package and initialize the `CMAKE_THREAD_LIBS_INIT` CMake variable, however this does not seem to be true for all platforms.

client/client-multi/client-cpp/CMakeLists.txt
RostakaGmfun: Yep, the `find_package(Threads)` fails if the project does not contain the C language.

vadimol (approved):

rasendubi (approved):

PR #1263, opened 2016-11-14
KAA-1441 Fix The ThreadPoolTest logic flow

ghost:
Reorganize test commands to avoid race condition.


vadimol (changes requested):
Please, remove commits from other developers

PR #1272, opened 2016-11-15
KAA-1585: fix email validation while update account profile

Serega290696:


server/node/src/main/java/org/kaaproject/kaa/server/admin/client/util/Utils.java
```diff
@@ -174,14 +176,15 @@ public static String millisecondsToDateTimeString(long millis) {
...
    */
   public static boolean validateEmail(String mail) {
-    boolean result = false;
-    if (mail != null && mail.length() != 0) {
-      if (mail.indexOf('@') != INCORRECT_IDX
-          && mail.indexOf('.') != INCORRECT_IDX) {
-        result = true;
-      }
+    if (mail == null || mail.length() == 0) {
+      return false;
```
Faradey27, line 183: looks good, 
but maybe in future we should return more complex error object, like

if email valid - return null
if email invalid return

{
  error: true,
  errorNumber: Integer,
  errorText: String
}

where errorText will have detailed explanation what was wrong in email, or something like

"Email could not be empty string"

this will help users better understand what we want from them

Faradey27:
lgtm

server/node/src/main/java/org/kaaproject/kaa/server/admin/client/util/Utils.java
```diff
@@ -174,14 +176,15 @@ public static String millisecondsToDateTimeString(long millis) {
...
     }
-    return result;
+    Pattern pattern = Pattern
```
garniy: it should be static final field

server/node/src/main/java/org/kaaproject/kaa/server/admin/client/util/Utils.java
Serega290696: Agree. I created a ticket 'KAA-1625'

server/node/src/main/java/org/kaaproject/kaa/server/admin/client/util/Utils.java
```diff
@@ -174,14 +176,15 @@ public static String millisecondsToDateTimeString(long millis) {
...
+        .compile(
+            "^[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,6}$", Pattern.CASE_INSENSITIVE
```
rasendubi: Kaa does not accept my local email `alexey@localhost`. And it does not accept my work mail address either: `ashmalko@mycompany.solutions`! I can't use Kaa!! Please help ASAP

rasendubi (changes requested):

server/node/src/main/java/org/kaaproject/kaa/server/admin/client/util/Utils.java
ohamov: My email doesn't work either: `"()<>[]:,;@\\\"!#$%&'-/=?^_`{}| ~.a"@gmail.com` :cry:
Could you please fix it? I want to try Kaa but the issue really blocks me.

server/node/src/main/java/org/kaaproject/kaa/server/admin/client/mvp/view/application/ApplicationViewImpl.java
```diff
@@ -74,9 +74,13 @@ protected void initDetailsTable() {
     applicationName.setFocus(true);
 
     if (!create) {
-      applicationToken = new KaaAdminSizedTextBox(DEFAULT_TEXTBOX_SIZE * 2, editable);
+//      applicationToken = new KaaAdminSizedTextBox(DEFAULT_TEXTBOX_SIZE * 2, editable);
+//      applicationToken.setWidth("100%");
+//      applicationToken.setEnabled(false);
```
rasendubi, line 79: Commented out code.

server/node/src/main/java/org/kaaproject/kaa/server/admin/client/util/Utils.java
```diff
@@ -80,9 +85,9 @@ public static void handleException(Throwable caught,
   /**
    * Exception handler.
    *
-   * @param caught                  the Throwable
-   * @param hasErrorMessage         the has error message
-   * @param errorMessageCustomizer  the error message customizer
+   * @param caught                 the Throwable
+   * @param hasErrorMessage        the has error message
+   * @param errorMessageCustomizer the error message customizer
```
rasendubi, line 90: This javadoc is not helpful. Could you please remove or expand it?

rasendubi:
@sergio-dev, please, leave a comment why you're closing.

garniy:
@rasendubi, it was replaced this [PR](https://github.com/kaaproject/kaa/pull/1308)

PR #1274, opened 2016-11-15
KAA-1567: Bad displaying of the Application details

Serega290696:


garniy (changes requested):
"20 of 510 max characters should be removed under the filed."
Please fix it

PR #1275, opened 2016-11-16
KAA-1497:updated cmake version in C++ SDK docs

yverzun:


yverzun:
closed, because contain commit from another ticket

PR #1296, opened 2016-11-17
KAA-1509: Your first Kaa application - fixed.

hunch7:
'Administration' changed to 'Administration UI'.

PR #1297, opened 2016-11-17
KAA-1509: Your first Kaa application - fixed.

hunch7:
'Administration' changed to 'Administration UI'.


vadimol (changes requested):
See comment in #1296

vadimol (approved):

PR #1303, opened 2016-11-18
KAA-1630:  GPIOdemo C client do not send event response

AlexShev76:


kaa-leeroy:

Slash32:
@kaa-leeroy add to whitelist

Slash32 (approved):
/cc @rasendubi this needs to be merged to master as well.

vadimol (dismissed):
This is a workaround. 
The Jira ticket has been created http://jira.kaaproject.org/browse/KAA-1630. 
The bug will be investigated and fixed in the scope of this ticket.

@AlexShev76 , please open the PR against master branch

rasendubi:
@AlexShev76 could you please elaborate on why the change is needed? (I have nothing against it, just trying to understand.)

Slash32:
@rasendubi it looks like the SDK and demo modules are somehow built with different headers that provide different size to size_t. Thus, the 64-bit value upon function call is passed as 64-bit and read out within the function as 32-bit dropping the lower 32-bit part, etc.

Slash32:
@kaa-leeroy test this please

rasendubi:
@Slash32 how casting values before calling the function helps? If the function implementation and invocation use different types, that's UB. (Casting values before invocation won't help.)

rasendubi (changes requested):
Now I don't trust the patch and would like this not merged until I get a good explanation of how this works.

vadimol:
Closing this PR until the proper fix will be provided.

PR #1313, opened 2016-11-21
KAA-1632: Updated formatting style for names of files and folders.

hunch7:


vadimol (approved):
Backport for #1315
Travis fail is not related.

PR #1324, opened 2016-11-23
APP-478: [C] Include header for using send_event function

Dubland:
/cc @vadimol @Slash32

### DIFF
--- a/client/client-multi/client-c/templates/event/kaa_event_families.cvm
+++ b/client/client-multi/client-c/templates/event/kaa_event_families.cvm
@@ -21,6 +21,7 @@
 
 # include "kaa_common.h"
 # include "kaa_event.h"
+# include "kaa_event_private.h"
 # include "kaa_error.h"
 # include "utilities/kaa_mem.h"
 


client/client-multi/client-c/src/extensions/event/kaa_event.h
```diff
@@ -101,6 +101,25 @@ kaa_error_t kaa_event_remove_transaction(kaa_event_manager_t *self, kaa_event_bl
 
 const char *kaa_find_class_family_name(const char *fqn);
 
+/*
+ * @brief Sends raw event
+ *
+ * It is not recommended to use this function directly. Instead you should use
```
ohamov: What about this comment?

Dubland: We do not use this function directly, but it is called in generated files.

vadimol: This looks odd.

ohamov (changes requested):
That's not the right solution. As the comment states, the function should not be used directly and it is not a part of the interface. The function declaration resides in `kaa_event_private.h`. Just add the necessary include to `kaa_event_families.cvm`.

vadimol (approved):

ohamov (approved):

vadimol:
Jenkins failure is not related.

PR #1349, opened 2016-11-30
KAA-1599: Updated Nix channel to nixos-16.09

yverzun:


rasendubi:
Travis has failed.

rasendubi:
Looks good. Now it's time to tackle other notes from the Jira ticket:
> Things to consider:
> - 16.09 has avro-c++
> - all current fixes and most patches (except for astyle) are present in 16.09 channel
> - 16.09 has minimal wine configuration, which is good to use for cc3200 SDK

Which means (respectively):
- remove avro-c++ files
- remove cmocka
- replace `wine` with `wineMinimal` for cc3200 sdk

rasendubi:
cmocka is not removed.

rasendubi:
macOS build fails because Hydra doesn't have pre-built binaries for cmake yet. Could you please disable the build with another commit?

rasendubi:
macOS build hasn't finished yet, but it's now allowed to fail, so I'm merging.

PR #1368, opened 2016-12-08
KAA-1561: Links fixed upon QA review. [master]

hunch7:

PR #1372, opened 2016-12-08
KAA-1561: Links fixed upon QA review. [release-0.10]

hunch7:
Done as per https://github.com/kaaproject/kaa/pull/1368


vadimol (approved):
Jenkins fail is not related

PR #1400, opened 2017-01-03
KAA-1658: [C++] Incorrect union for DEFAULT_WORKING_DIR

Dubland:
/cc @Slash32

### DIFF
--- a/client/client-multi/client-cpp/impl/KaaClientProperties.cpp
+++ b/client/client-multi/client-cpp/impl/KaaClientProperties.cpp
@@ -25,9 +25,9 @@
 namespace kaa {
 
 #if __GNUC__
-    static const char FILE_SEPARATOR = '/';
+    static const char DIRECTORY_SEPARATOR = '/';
 #else
-    static const char FILE_SEPARATOR = '\\';
+    static const char DIRECTORY_SEPARATOR = '\\';
 #endif
 
 const std::string KaaClientProperties::PROP_WORKING_DIR = "kaa.work_dir";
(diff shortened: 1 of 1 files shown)


client/client-multi/client-cpp/impl/KaaClientProperties.cpp
```diff
@@ -95,12 +95,8 @@ void KaaClientProperties::setWorkingDirectoryPath(const std::string& path)
 {
     checkEmptyness(path, "Empty value of working directory path");
 
-    if (path[path.size() - 1] != FILE_SEPARATOR) {
-        std::string editedPath = path + &FILE_SEPARATOR;
+        std::string editedPath = path + FILE_SEPARATOR;
```
rasendubi: Code style.

rasendubi: It does not handle the case when `path` already has a trailing separator.

RostakaGmfun: Use [`std::basic_string::compare`](http://en.cppreference.com/w/cpp/string/basic_string/compare)

client/client-multi/client-cpp/impl/KaaClientProperties.cpp
```diff
@@ -25,9 +25,9 @@
 namespace kaa {
 
 #if __GNUC__
-    static const char FILE_SEPARATOR = '/';
+    static const char *FILE_SEPARATOR = "/";
```
ohamov: Let's move this property to `KaaClientProperties` and use `std::string` instead of `char*`

ohamov: Also, rename to `directory_separator` please.

client/client-multi/client-cpp/impl/KaaClientProperties.cpp
```diff
@@ -95,8 +95,8 @@ void KaaClientProperties::setWorkingDirectoryPath(const std::string& path)
 {
     checkEmptyness(path, "Empty value of working directory path");
 
-    if (path[path.size() - 1] != FILE_SEPARATOR) {
-        std::string editedPath = path + &FILE_SEPARATOR;
+    if (path.compare(path.length()-DIRECTORY_SEPARATOR.length(), DIRECTORY_SEPARATOR.length(), DIRECTORY_SEPARATOR) != 0) {
```
GamovCoder: I think we can eliminate this check, the path should be normalized.

ohamov (approved):

client/client-multi/client-cpp/impl/KaaClientProperties.cpp
```diff
@@ -25,9 +25,9 @@
 namespace kaa {
 
 #if __GNUC__
-    static const char FILE_SEPARATOR = '/';
+    static const char DIRECTORY_SEPARATOR = '/';
```
rasendubi, line 28: `PATH_SEPARATOR`?

rasendubi (approved):

PR #1490, opened 2017-02-23
CI-26: Add ability to disable `jekyll serve` in tst_deploy.sh

RostakaGmfun:
/cc @balldir


RostakaGmfun:
@kaa-leeroy test this please

balldir (approved):

Slash32 (changes requested):
This file originates from Jekver project now and thus should be modified there. Otherwise it will be replaced during documentation compilation phase leading to unexpected behavior.

Slash32 (approved):
Approved upon local discussion.

PR #1495, opened 2017-03-15
EVA-105: Dependency management versions fixed

odovhai:


odovhai:
@kaa-leeroy test this please

